PR #73, opened 2026-06-17
fix(codex): fix false compact alert from cumulative token count

jomeswang:
## Summary

- **Root cause**: `thread/tokenUsage/updated` carries two `TokenUsageBreakdown` objects — `last` (most-recent API call) and `total` (cumulative thread totals). Commit 1a31653 swapped the preference to `total.totalTokens`, causing Tutti to compare a growing cumulative sum against the per-request `modelContextWindow`. After ~10 tool calls of ~27 K tokens each, the cumulative reaches 270 K, exceeds the 258 K window, and fires a false "context almost full" compact alert — even though each individual request was only ~8% full.

- **Fix**: Use `last.inputTokens` (exact input tokens sent to model = true context fill) with `last.totalTokens` as fallback, and `total.totalTokens` only if `last` is absent entirely. Verified against the official Codex App Server JSON schema (`codex app-server generate-json-schema`) and confirmed live with a test script.

- **Also**: Surfaces Codex rate-limit quotas at session Start/Resume so the UI shows the correct value immediately instead of "unavailable" until the first push notification. Removes a debug artifact (`agentDebugLog`) that wrote to a hardcoded absolute path on every model-switch.

## Evidence

Live verification against real Codex App Server (3 turns, default model):

```
turn 1  last.input=22195  last.total=22217  cumul.total=22217   fill(correct)=8.6%   fill(wrong)=8.6%
turn 2  last.input=20467  last.total=20472  cumul.total=42689   fill(correct)=7.9%   fill(wrong)=16.5%
turn 3  last.input=20486  last.total=20492  cumul.total=63181   fill(correct)=7.9%   fill(wrong)=24.5%
```

`cumul.total` doubles each turn; `last.input` stays stable at ~20 K. After ~12 turns it would have exceeded the 258 K window.

## Test plan

- [ ] `go test ./packages/agent/daemon/runtime/...` — existing + new tests pass
- [ ] `TestCodexAppServerAdapterApplyTokenUsagePrefersInputTokens` — verifies `last.inputTokens` is preferred
- [ ] `TestCodexAppServerAdapterApplyTokenUsageFallsBackToLastTotalTokens` — verifies fallback when `inputTokens` absent
- [ ] `TestCodexAppServerAdapterApplyTokenUsageNoCumulativeFalsePositive` — simulates 10 calls, asserts `usedTokens` never exceeds window
- [ ] Run `node tools/scripts/verify-codex-token-usage.mjs` against live Codex App Server to confirm semantics

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


devRickyyy:
Closing to re-create under the correct account (devRickyyy).

PR #261, opened 2026-06-22
Dedupe Codex reply snapshots

devRickyyy:
## Summary
- dedupes Codex reply snapshots during ACP turn normalization and message projection
- prevents repeated assistant reply rendering when the same final snapshot is seen more than once
- adds daemon and renderer projection coverage

## Validation
- `git status --short --branch` clean
- targeted tests were added by the fixing agent
- pre-push `pnpm check:full` was blocked by existing Go test failures in `services/tuttid/service/agentstatus` and `services/tuttid/types`


devRickyyy:
关闭，不合入。分支保留。

PR #317, opened 2026-06-24
fix(agent): bundle Claude Code ACP adapter for offline install

devRickyyy:
## Problem

Installing **Claude Code** could fail during onboarding with:

```
Installation failed — sh: claude-agent-acp: command not found
```

The ACP adapter (`@agentclientprotocol/claude-agent-acp`) is installed on the
user's machine at runtime via `npm install` from the public registry. On
slow/blocked networks the install hangs past the 5‑minute timeout and is killed,
leaving the adapter binary absent; the subsequent launch then reports
`command not found`. (Diagnosed from a user log bundle: `installerKind=external_agent_registry_npm`
→ `error="context deadline exceeded"` at exactly 5 min.)

## Approach

Ship the ACP bridge **with the desktop app** so Claude Code works without a
runtime npm install — mirroring the existing `chrome-devtools-mcp` vendoring:

- **Build time** (`vendor-claude-acp.mjs`): `npm install` the pinned bridge, run
  the existing fast‑mode/`/goal` codemod against it, and ship it via
  electron-builder `extraResources`.
- **Daemon** (`tuttidManager.ts`): in packaged builds, point the daemon at the
  vendored entry with `TUTTI_CLAUDE_ACP_ENTRY_PATH` (no-op in dev / operator
  override, mirroring `resolveBrowserMcpDaemonEnv`).
- **Resolution** (`claude_acp_bundled.go`, `provider_resolution.go`): when the
  bundle is present, `resolveProviderSpec` runs it directly with the managed Node
  runtime and **never** falls back to the registry/npm path.

## Issues found & fixed while verifying (3 rounds of on-device + CI testing)

1. **node_modules stripped from the package.** electron-builder hard-excludes a
   `node_modules` dir at the root of an `extraResources` `from` dir
   (`app-builder-lib util/filter.js`: `relative === "node_modules"` → excluded),
   but keeps a *nested* one. → Install the vendored tree under `bridge/`.
   *(Note: `build/browser-mcp` has the same latent bug — its vendored tree is
   also stripped, so packaged browser-use silently falls back to `npx`. Not
   fixed here; see follow-ups.)*
2. **216 MB bundled CLI broke the universal build.** The SDK pulls in
   `@anthropic-ai/claude-agent-sdk-<os>-<arch>` (the full Claude Code CLI as a
   platform-specific native dep); only the build host's arch is installed, so
   `@electron/universal` can't merge it. → Prune those native packages
   (`build/claude-acp`: ~216 MB → ~43 MB, no Mach-O left) and point the bridge at
   Tutti's system-managed `claude` via `CLAUDE_CODE_EXECUTABLE` (the bridge's
   `claudeCliPath()` honors it first).
3. **Startup race fell back to npm + version mismatch.** At the first resolution
   the managed Node runtime was still materializing, so the short circuit fell
   through to the registry; the npm install ran, then the recheck reported the
   bundled `0.46.0` against the registry's required `0.50.0` →
   `provider adapter is still unavailable after install`. → Once shipped the
   bundle is authoritative: always return the bundled spec (never the registry)
   and keep `AdapterInstall` empty so no npm install can ever be selected; if the
   runtime isn't ready yet, leave the command empty and self-heal on the next
   resolve.

## Verification

- Go (`agentstatus`) + desktop (`tuttidManager`) unit/regression suites green;
  gofmt/vet clean.
- Vendored bridge smoke-tested: answers ACP `initialize` with the bundled CLI
  pruned and `CLAUDE_CODE_EXECUTABLE` set.
- Unsigned dry-run packaging (arm64/x64/universal) verified: shipped app contains
  `bin/claude-acp/bridge/node_modules/.../dist/index.js`, no Mach-O binaries,
  fast-mode patch applied.
- **On-device (rc.274):** enabling Claude Code shows **zero** install steps in
  `tuttid.log`, the bundled `claude-agent-acp` adapter starts, authenticates, and
  runs sessions; SDK init loads the user's own plugins/skills (proving it uses
  the system `claude`).

## Out of scope / follow-ups

- `browser-mcp` has the same `node_modules`-stripping bug (packaged browser-use
  falls back to `npx`); fixable with the same nesting.
- Bundled bridge is pinned to `0.46.0` (registry currently serves `0.50.0`);
  bump separately once patch anchors are confirmed.
- First-run still needs network to download the managed Node runtime (out of
  scope; the eliminated failure was the npm adapter install).

## Summary by CodeRabbit

* **New Features**
  * The desktop app now ships with a pre-packaged Claude ACP bridge, enabling agent integration to work reliably in packaged/offline environments without downloading at runtime.

* **Tests**
  * Added unit and service-level coverage to verify bundled bridge selection, correct environment setup, and expected command-generation behavior, including proper fallback when the bundled entry is unavailable.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -27,6 +27,10 @@
       {
         "from": "build/browser-mcp",
         "to": "bin/browser-mcp"
+      },
+      {
+        "from": "build/claude-acp",
+        "to": "bin/claude-acp"
       }
     ],
     "publish": [
--- a/apps/desktop/scripts/vendor-claude-acp.mjs
+++ b/apps/desktop/scripts/vendor-claude-acp.mjs
@@ -22,7 +22,14 @@ const CLAUDE_ACP_VERSION = "0.46.0";
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const desktopDir = join(__dirname, "..");
 const repoRoot = join(desktopDir, "..", "..");
-const outDir = join(desktopDir, "build", "claude-acp");
+// stageDir is the extraResources `from` root (shipped to Resources/bin/claude-acp).
+// We install one level BELOW it, under bridge/, because electron-builder's file
+// matcher hard-excludes a `node_modules` directory that sits at the ROOT of an
+// extraResources `from` dir (app-builder-lib util/filter.js: `relative ===
+// "node_modules"` -> excluded) but KEEPS a nested one. Without this nesting the
+// vendored dependency tree is silently stripped from the packaged app.
+const stageDir = join(desktopDir, "build", "claude-acp");
+const outDir = join(stageDir, "bridge");
 const packageDistDir = join(
   "node_modules",
   "@agentclientprotocol",
@@ -47,7 +54,7 @@ function log(msg) {
   process.stderr.write(`[vendor-claude-acp] ${msg}\n`);
 }
 
-rmSync(outDir, { recursive: true, force: true });
+rmSync(stageDir, { recursive: true, force: true });
 mkdirSync(outDir, { recursive: true });
 
 // A minimal package.json so `npm install` materializes a self-contained tree.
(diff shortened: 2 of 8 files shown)


devRickyyy:
Superseded by #321. The product need is reliable install on slow/blocked networks, not airgapped-offline, so we drop the app-bundling approach (and all its electron-builder/universal/race workarounds) in favor of pointing the adapter's npm install at a reliable registry mirror.

PR #465, opened 2026-06-28
feat(agent-gui): enter conversation immediately on first message

devRickyyy:
## What

Sending the first message in a new chat now enters the conversation surface **immediately** and shows the user's own message, instead of waiting through the activation loading period before entering.

## Why

Previously the first-message create kept the user on the home composer until the backend session-creation round-trip resolved, only then switching into the conversation. This made the first send feel sluggish. The desired behavior is: send → instantly in the conversation seeing your message.

A prior PRD conflated "immediate entry" with showing a "正在连接会话…" (reconnecting) banner. This PR decouples them: immediate entry does **not** require the banner.

## How

- **Optimistic commit moved into the synchronous submission block** of `startConversation`. On submit we set the transient conversation, record the optimistic user message (keyed by `clientSubmitId`), clear the session draft, and flip `activeConversationId`/`isComposerHome`/`intent` to the new session — all before awaiting `activation.activate`. The home draft is intentionally left intact so a failed activation can revert with the user's input preserved.
- **`clientSubmitId` reconciliation reused** (no new mechanism): the optimistic turn uses `pending:<clientSubmitId>` and is retargeted to the real turn when the durable message lands, exactly as before — only the timing of the optimistic commit changed.
- **Banner suppressed during initial create**: `sessionChrome.recovery` only shows the `activating` banner when the active conversation is *not* the in-flight first-message create (`startingConversationIdRef.current !== activeConversationId`). Activation-in-flight on a brand-new session is not a recovery event.
- **Failure reverts to home**: the `.then` `activationFailed` branch and `.catch` undo the optimistic entry — back to the home composer with the original draft preserved and the error surfaced there (transient/overlay cleared, `activeConversationId=null`).
- **`pendingHomeErrorRef`**: the existing `activeConversationId`-null effect cleared `detailError` on any home transition, which would wipe the just-set failure error during the revert. A small ref stashes the error so the effect surfaces it on the home composer instead of clearing it. The ref is only set when actually reverting (`isCurrentConversation`), so it can't leak into unrelated home transitions.

## Verification

- `pnpm typecheck` (agent-gui) ✅
- `pnpm oxlint` on changed files ✅ (exit 0)
- `pnpm test` (agent-gui): 105 files / 1677 passed / 1 skipped ✅
- Updated the 4 tests that asserted the old "stay on composer" behavior; added a banner-suppression test asserting `activeLiveState === "activating"` with `sessionChrome.recovery === null` during a pending first-message create.

## Summary by CodeRabbit

* **Bug Fixes**
  * First-message conversation creation now switches to the conversation view immediately, showing the optimistic message while activation is still pending.
  * Suppresses the reconnecting banner during the initial pending activation for the optimistic first-message flow.
  * If activation or creation fails, the app returns to the home composer with the correct draft restored, clears creation busy/loading state, and keeps the error associated with the right conversation even after navigation changes.

* **Tests**
  * Expanded regression coverage for pending flows, navigation-away/drop behavior, and draft restoration after activation rejection.

### DIFF
--- a/packages/agent/gui/agent-gui/agentGuiNode/controller/useAgentGUINodeController.spec.tsx
+++ b/packages/agent/gui/agent-gui/agentGuiNode/controller/useAgentGUINodeController.spec.tsx
@@ -1255,6 +1255,163 @@ describe("useAgentGUINodeController", () => {
     expect(result.current.viewModel.sessionChrome.recovery).toBeNull();
   });
 
+  it("drops the optimistic entry and keeps the user put when they navigate away before a first-message create resolves", async () => {
+    let resolveActivate:
+      | ((result: AgentHostActivateAgentSessionResult) => void)
+      | undefined;
+    let createdId = "";
+    const activate = vi.fn((input: AgentHostActivateAgentSessionInput) => {
+      if (input.mode === "new") {
+        createdId = input.agentSessionId;
+        return new Promise<AgentHostActivateAgentSessionResult>((resolve) => {
+          resolveActivate = resolve;
+        });
+      }
+      return Promise.resolve<AgentHostActivateAgentSessionResult>({
+        session: agentSession(input.agentSessionId),
+        activation: { mode: input.mode, status: "attached" }
+      });
+    });
+    installAgentHostApi({
+      list: vi.fn(async () => snapshotWithSession("session-1")),
+      listSessionTimeline: vi.fn(async () => ({ timelineItems: [] })),
+      subscribeEvents: vi.fn(() => vi.fn()),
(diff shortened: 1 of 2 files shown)


devRickyyy:
这个分支已经stale（main在这之后经历了乐观/reconcile重写，version-domain不变量已经变了），改为在当前main上重新实现同样的意图，保留了这个PR原本的设计（pendingHomeErrorRef、isCurrentConversation错误归属、失败回退逻辑），并额外确认了Codex和Claude Code都适用。新PR见后续链接。分支保留。

PR #471, opened 2026-06-28
fix(agent-gui): prevent images from remaining in composer after send

### LINKED ISSUE #430: Sent images remain in the composer after the message is sent
## Summary
After sending a message with images, the images can remain visible in the composer for a while.

## Steps to reproduce
1. Open Tutti desktop.
2. Navigate to Agent GUI.

## Expected behavior

## Actual behavior
After sending a message with images, the images can remain visible in the composer for a while.

## Affected area
Agent GUI

## Platform
Desktop app

## Additional context
- Priority: P2
- Affected area: Agent GUI
- Platform: Desktop app
- Public screenshots/videos:
  - Screenshot 1
- Private logs: omitted from public issue mirror

Shiyao-Huang:
## Summary

- Fixes #430 — Sent images remain in the composer after the message is sent

## Root Cause

When images are pasted into the composer, they start async uploads (`uploadPromptContent`). The `.then()` and `.catch()` callbacks update `draftImagesRef` and call `onDraftContentChange()` to reflect upload progress. After the user clicks send, `submitCurrentPrompt` clears the refs and draft state synchronously. However, a pending upload completion callback can fire after submit, read the ref (which may have been re-populated by the `draftImages → draftImagesRef` sync effect during an intermediate render), and call `onDraftContentChange` with image data — re-adding the images to the composer.

## Fix

Add a `submittedRef` guard flag that is set to `true` when the user submits the prompt. The upload completion callbacks (`.then()` and `.catch()`) now check this guard and skip the draft content update if submit has already occurred. The guard is reset to `false` when new images are added to the draft.

## Test plan

- [x] Code passes pre-commit hooks (oxlint, oxfmt)
- [ ] Manual: paste an image into composer → wait for upload to complete → send → images clear immediately
- [ ] Manual: paste an image → send quickly before upload completes → submit is blocked (existing behavior)
- [ ] Manual: paste an image → send → paste another image in new message → works normally

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved composer behavior so image uploads no longer update or overwrite draft content after a prompt has already been submitted.
  - Reduced the chance of confusing draft image states when sending a message while uploads are still finishing.

### DIFF
--- a/packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.tsx
+++ b/packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.tsx
@@ -813,6 +813,7 @@ export function AgentComposer({
   const draftPromptRef = useRef(draftPrompt);
   const draftImagesRef = useRef<AgentComposerDraftImage[]>(draftImages);
   const draftFilesRef = useRef<AgentComposerDraftFile[]>(draftFiles);
+  const submittedRef = useRef(false);
   const promptTipRef = useRef<HTMLSpanElement | null>(null);
   const mentionControllerRef = useRef<AgentMentionSearchController | null>(
     null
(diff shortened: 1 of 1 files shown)


Shiyao-Huang:
Closing as duplicate of #473 which has the complete fix.

PR #478, opened 2026-06-28
fix(agent-gui): reset local submitting flag on turn settled

### LINKED ISSUE #428: Completed conversations briefly flash queued state before sending each message
## Summary
After a conversation has completed, sending another message briefly flashes a queued state before the message is sent.

## Steps to reproduce
1. Open Tutti desktop.
2. Navigate to Agent GUI.

## Expected behavior

## Actual behavior
After a conversation has completed, sending another message briefly flashes a queued state before the message is sent.

## Affected area
Agent GUI

## Platform
Desktop app

## Additional context
- Priority: P2
- Affected area: Agent GUI
- Platform: Desktop app
- Public screenshots/videos:
  - [Recording 1 (.mov, 8.4 MB)](https://github.com/tutti-os/tutti-issue-assets/releases/download/feishu-p2-bugs-2026-06-27/issue-428-1-EY2FbYD1WomGDMxn8CscDgnbnyo.mov)

Shiyao-Huang:
## Problem

After a conversation has completed, sending another message briefly flashes a queued state before the message is actually sent.

Fixes #428

## Root Cause

When a conversation completes, the server sends a turn-settled event that clears the pending turn ID. However, the local submitting flag `localIsSubmitting` is only reset in the `.finally()` callback of the original `sendInput` promise, which may run slightly later. During this gap, `shouldQueuePromptLocally` returns `true` (because `isSubmitting` is still true), causing the next message to briefly enter the queued state.

## Solution

When clearing the pending submitted turn due to the conversation transitioning to a non-busy status, also reset `localIsSubmitting` immediately. This closes the race window between the server completion event and the promise cleanup.

## Changes

- `packages/agent/gui/agent-gui/agentGuiNode/controller/useAgentGUINodeController.ts`: Reset `localIsSubmitting` when the turn is cleared due to non-busy conversation status.

## Changeset

- @tutti-os/agent-gui: patch

### DIFF
--- a/.changeset/reset-submitting-on-turn-settled.md
+++ b/.changeset/reset-submitting-on-turn-settled.md
@@ -1,0 +1,5 @@
+---
+"@tutti-os/agent-gui": patch
+---
+
+Reset the local submitting flag as soon as the server confirms a turn has settled, rather than waiting for the original send request promise to resolve. Fixes #428 where completed conversations briefly flashed a queued state when sending the next message.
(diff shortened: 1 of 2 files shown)


Shiyao-Huang:
Superseded by #494.

PR #527, opened 2026-06-29
fix(agent-gui): stabilize message display ordering

rv4no:
## Summary

- Order AgentGUI durable message projection by display time derived from `startedAtUnixMs` first, with existing version/id fallbacks for ties.
- Use the same start-time-first value as the projected timeline item timestamp so late-completing tool calls stay before the final assistant response in old and new conversations.
- Scope the transient processing row identity to the latest turn so a new running turn does not reuse the previous turn's loading row identity.
- Add regression coverage for both late-completing tool calls and turn-scoped processing rows.

## Verification

- `pnpm --filter @tutti-os/agent-gui exec oxfmt --check packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.spec.ts`
- `pnpm --filter @tutti-os/agent-gui exec vitest run shared/agentConversation/projection/workspaceAgentMessageProjection.spec.ts`
- `pnpm --filter @tutti-os/agent-gui exec oxfmt --check shared/agentConversation/projection/agentProcessingProjection.ts shared/agentConversation/projection/agentConversationProjection.spec.ts`
- `pnpm --filter @tutti-os/agent-gui exec vitest run shared/agentConversation/projection/agentConversationProjection.spec.ts`
- `pnpm check:full`

## Checklist

- [x] I updated documentation when behavior, setup, or contributor workflow changed. No durable docs were needed for these projection-only display fixes; behavior is covered by regression tests.
- [x] I updated all README/CONTRIBUTING language variants when changing their English source. Not applicable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation message ordering so tool-call rows and the final assistant response appear in the correct sequence, including when tool completion is delayed.
  * Updated the timestamp logic used for display ordering to better match what users see in the timeline.
  * Fixed the “processing” indicator so it’s tied to the latest turn rather than a generic session marker.
* **Tests**
  * Added test coverage for ordering with late-completed tool calls.
  * Added test coverage ensuring the processing indicator targets the most recent turn.

### DIFF
--- a/packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
+++ b/packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
@@ -266,6 +266,10 @@ function compareMessagesByDisplayOrder(
   left: WorkspaceAgentActivityMessage,
   right: WorkspaceAgentActivityMessage
 ): number {
+  // This comparator decides the rendered message position. startedAt is the
+  // start time for long-running items, occurredAt is the append time for plain
+  // messages, and completedAt is only a last fallback; version/id are stable
+  // tie-breakers when two messages share the same display time.
   return (
     messageDisplayOrderTime(left) - messageDisplayOrderTime(right) ||
     normalizedPositiveNumber(left.version) -
--- a/packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
+++ b/packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
@@ -49,10 +49,7 @@ export function projectWorkspaceAgentMessagesToTimelineItems(
     const seq = index + 1;
     const eventId = message.messageId.trim() || `message:${id}`;
     const turnId = message.turnId?.trim() || undefined;
-    const occurredAtUnixMs =
-      message.occurredAtUnixMs ??
-      message.completedAtUnixMs ??
-      message.startedAtUnixMs;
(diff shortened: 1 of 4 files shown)


packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
```diff
@@ -292,10 +289,12 @@ function normalizedPositiveNumber(value: number | undefined): number {
     : 0;
 }
 
-function messageOrderTime(message: WorkspaceAgentActivityMessage): number {
+function messageDisplayOrderTime(
+  message: WorkspaceAgentActivityMessage
+): number {
   return (
-    normalizedPositiveNumber(message.occurredAtUnixMs) ||
     normalizedPositiveNumber(message.startedAtUnixMs) ||
```
rv4no, line 300: occur 和 start 分别对应什么时候

packages/agent/gui/shared/agentConversation/projection/workspaceAgentMessageProjection.ts
```diff
@@ -270,9 +267,9 @@ function compareMessagesByDisplayOrder(
   right: WorkspaceAgentActivityMessage
 ): number {
   return (
+    messageDisplayOrderTime(left) - messageDisplayOrderTime(right) ||
```
rv4no, line 274: 这里为啥前置了

PR #621, opened 2026-07-01
fix(activity-core): preserve failed session status over non-terminal turn status

Shiyao-Huang:
## Summary
- In `selectSessionDisplayStatuses`, when a session was marked "failed", the display status was overridden by `latestTurnStatus` regardless of whether that turn status was terminal. This caused error sessions to appear as "working" when the latest message still had a running/streaming status.
- Only override the session failure with `latestTurnStatus` when it is itself a terminal display status (completed, failed, canceled, idle). This ensures error sessions show as failed in the message center until a terminal turn status arrives.
- Fixes the pre-existing test failure: "counts error message-center sessions as failed, not completed"

## Verification

- All 13 selector tests pass (`node --test src/selectors.test.ts`)
- All 33 message center model tests pass (including previously failing test)
- Pre-commit hooks pass (oxfmt, oxlint, boundary checks)
- `pnpm check:full` passes

## Checklist

- [ ] I updated documentation when behavior, setup, or contributor workflow changed.
- [ ] I updated all README/CONTRIBUTING language variants when changing their English source.

## AI Disclosure
- Code was written with assistance from Claude (Anthropic). The fix was reviewed and verified by running the existing test suite and TypeScript type checking.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session status display so the latest turn status only overrides the session’s main status when it represents a final outcome.
  * Prevents in-progress or intermediate statuses from replacing the normalized session status, making activity state more accurate and consistent.

### DIFF
--- a/packages/agent/activity-core/src/selectors.ts
+++ b/packages/agent/activity-core/src/selectors.ts
@@ -50,7 +50,10 @@ export function selectSessionDisplayStatuses(
       return [
         session.agentSessionId,
         sessionStatus === "failed"
-          ? (latestTurnStatus ?? sessionStatus)
+          ? latestTurnStatus != null &&
+            isTerminalDisplayStatus(latestTurnStatus)
+            ? latestTurnStatus
+            : sessionStatus
           : sessionStatus
       ];
     })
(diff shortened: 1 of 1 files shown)


devRickyyy:
Thank you for the contribution and for calling attention to stale failure status handling.

The activity/status architecture has changed substantially since this PR was opened. The old `selectSessionDisplayStatuses` path modified here no longer exists; #686, #1058, and #1093 moved status authority to the canonical interaction/turn model (`pendingInteractions`, `activeTurn.phase`, and `latestTurn.outcome`). In the current model, a newer running turn is intentionally allowed to override an older session-level failure. Applying this patch's semantics now would reintroduce a sticky stale-failure badge.

Because the target code and the intended status semantics have both been superseded, we are closing this PR. If a failed badge still persists incorrectly on current `main`, a fresh report with the session lifecycle tuple or diagnostic bundle would be useful; the fix should then be made at the daemon/adapter turn-authority boundary rather than by restoring the removed selector.

Issue #423, opened 2026-06-27
Message Center renders stray asterisks around messages


devRickyyy:
## Summary
Message Center can render messages with stray double asterisks before and after the content.

## Steps to reproduce
1. Open Tutti desktop.
2. Navigate to Agent status/message panel.
3. Perform the workflow described in the summary.
4. Observe the incorrect behavior.

## Expected behavior
Agent status/message panel should complete the workflow without the problem described above.

## Actual behavior
Message Center can render messages with stray double asterisks before and after the content.

## Affected area
Agent status/message panel

## Platform
Desktop app

## Additional context
- Priority: P2
- Affected area: Agent status/message panel
- Platform: Desktop app
- Public screenshots/videos:
  - [Recording 1 (.mov, 49.3 MB)](https://github.com/tutti-os/tutti-issue-assets/releases/download/feishu-p2-bugs-2026-06-27/issue-423-1-UdFubRvYBocSIcxdwZfcsklJn6e.mov)

PR #623, opened 2026-07-01
fix(agent-gui): strip markdown emphasis from message center titles and previews

Shiyao-Huang:
## Summary
- Strip markdown bold/italic markers (asterisks and underscores) from message center titles and stack preview text segments
- Restore `normalizeAgentTitleText` import that was removed during a prior refactor, causing raw markdown to leak into the message center UI
- Add `stripPreviewTextEmphasis` helper with targeted patterns for `**bold**`, `__bold__`, `*italic*`, `_italic_` on stack preview segments
- Fix summary fallback to normalize through `normalizeAgentTitleText` for consistent display

## Verification

- Added 11 unit tests in `agentTitleText.spec.ts` covering null/undefined, plain text, links, bold/italic markers (asterisk + underscore), bold-italic triple markers, multiplication sign preservation, whitespace collapsing, and combined patterns
- Pre-commit hooks pass (oxfmt, oxlint, boundary checks)
- `pnpm check:full` passes

## Checklist

- [ ] I updated documentation when behavior, setup, or contributor workflow changed.
- [ ] I updated all README/CONTRIBUTING language variants when changing their English source.

## AI Disclosure
- Code was written with assistance from Claude (Anthropic). The fix was reviewed and verified by running the existing test suite.

### DIFF
--- a/packages/agent/gui/agent-message-center/WorkspaceAgentMessageCenterCard.tsx
+++ b/packages/agent/gui/agent-message-center/WorkspaceAgentMessageCenterCard.tsx
@@ -35,6 +35,7 @@ import {
 } from "@tutti-os/ui-rich-text/core";
 import { getActiveUiLanguage, useTranslation } from "../i18n/index";
 import { formatAgentSessionMentionText } from "../shared/utils/agentSessionMentionText";
+import { normalizeAgentTitleText } from "../shared/utils/agentTitleText";
 import { AgentInteractivePromptSurface } from "../shared/agentConversation/components/AgentInteractivePromptSurface";
 import { AgentMessageMarkdown } from "../shared/AgentMessageMarkdown";
 import { AgentVerticalScrollArea } from "../shared/AgentVerticalScrollArea";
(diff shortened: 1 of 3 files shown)


devRickyyy:
Thank you for the contribution and for tracing the raw Markdown markers shown in Message Center previews. The compact-preview problem tracked by #423 is still valid.

The ownership and rendering paths have changed since this PR was opened. Session-title canonicalization now belongs to the daemon, while compact Message Center preview extraction lives in a dedicated presentation helper. Applying a global emphasis-stripping regex in the renderer is no longer the right boundary, and the proposed pattern can also corrupt valid plain text such as `foo_bar_baz`, identifiers, or file paths.

We are closing the original PR because it is not safe to merge or mechanically rebase. A focused follow-up should convert Markdown to plain text only in the compact preview projection, preserve identifiers and paths, and cover cases such as `**bold**`, `_italic_`, `foo_bar_baz`, and file-path content. If product requirements also call for normalized session titles, that change should be made in the daemon title-normalization layer. The original report and investigation remain useful—thank you.

PR #689, opened 2026-07-04
fix(file-manager): support arrow-key drill-down in the file tree list

devRickyyy:
## Problem

In the file/artifact list panel (list layout mode), the current state does not support left/right (or up/down) arrow keys to navigate into a subfolder — pressing an arrow key does nothing, forcing users to click the disclosure triangle with a mouse. (Feishu tracker CPF5q8, P1)

## Root cause

File list rows in `packages/workspace/file-manager/src/ui/WorkspaceFileManagerPanels.tsx` were plain buttons with no keyboard handling beyond native Enter/Space activation. There was no `onKeyDown` wiring for `ArrowLeft`/`ArrowRight`/`ArrowUp`/`ArrowDown`, so directory drill-down (lazy-loading the next-level listing) and row-to-row navigation were only reachable via mouse click on the disclosure control.

## Fix

- Wired `ArrowRight`/`ArrowLeft` to the same `onToggleDirectoryExpanded` path the existing disclosure button already uses — this lazy-loads the folder's children only when the user actually navigates into it, and reuses the existing loading/empty/error feedback row rather than introducing new UI states.
- Wired `ArrowUp`/`ArrowDown` to move selection between the currently visible rows.
- Changes are in `packages/workspace/file-manager/src/ui/WorkspaceFileManagerPanels.tsx` and `packages/workspace/file-manager/src/ui/workspaceFileManagerVisibleTree.ts`.

## Tests

- `pnpm test` (`node --test`) in `packages/workspace/file-manager` — 141/141 passed, including new keyboard-navigation cases (`keyboard navigation finds adjacent, parent, and first-child rows across an expanded tree`, `keyboard navigation surfaces the loading/empty/error feedback row as the first child of an expanded directory`).
- `pnpm typecheck` in `packages/workspace/file-manager` — clean.

## Note on CI bypass

This branch was pushed with `git push --no-verify`, explicitly authorized by the user, because the local pre-push hook (`pnpm check:full`) deterministically hangs in an unrelated, pre-existing test: `TestServiceCreateResolvesProviderFromAgentTarget` in `services/tuttid/service/agent/service_test.go` polls forever in `pollComposerModelOptions` (`services/tuttid/service/agent/composer_live_model_discovery.go:135`) because the test's fake runtime double never populates `RuntimeContext` or marks the session `"failed"` — a latent gap since `ecf73863` ("require verified Claude composer models"). This diff contains **no Go changes** and cannot affect that test; the hang was independently reproduced in isolation (`go test ./service/agent/... -run TestServiceCreateResolvesProviderFromAgentTarget -timeout 60s -count=3`, hangs all 3 times) and confirmed non-flaky across multiple retries under varying system load. All other check:full lanes and the tests specific to this change were verified manually (above).

## Summary by CodeRabbit

* **New Features**
  * Added keyboard navigation in the workspace file manager for moving between files and folders with arrow keys.
  * Directory navigation now supports expanding/collapsing folders and moving into parent or first child items from the keyboard.

* **Bug Fixes**
  * Improved focus behavior so the selected item stays in sync with keyboard navigation and remains visible in the list.

### DIFF
--- a/packages/workspace/file-manager/src/ui/WorkspaceFileManagerPanels.tsx
+++ b/packages/workspace/file-manager/src/ui/WorkspaceFileManagerPanels.tsx
@@ -13,6 +13,7 @@ import type { WorkspaceFileManagerI18nRuntime } from "../i18n/workspaceFileManag
 import type {
   CSSProperties,
   DragEvent as ReactDragEvent,
+  KeyboardEvent as ReactKeyboardEvent,
   MouseEvent as ReactMouseEvent,
   PointerEvent as ReactPointerEvent,
   ReactElement,
@@ -49,7 +50,12 @@ import {
   type WorkspaceFileManagerArrangeMode
 } from "./workspaceFileManagerArrangeMode.ts";
 import type { WorkspaceFileManagerLayoutMode } from "./workspaceFileManagerLayoutMode.ts";
-import type { WorkspaceFileManagerVisibleTreeRow } from "./workspaceFileManagerVisibleTree.ts";
+import {
+  findWorkspaceFileManagerAdjacentEntryPath,
+  findWorkspaceFileManagerFirstChildRow,
+  findWorkspaceFileManagerParentEntryPath,
+  type WorkspaceFileManagerVisibleTreeRow
+} from "./workspaceFileManagerVisibleTree.ts";
 
 const workspaceFileManagerTableGridClassName =
   "grid-cols-[minmax(0,_1fr)_148px_96px]";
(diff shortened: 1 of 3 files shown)


devRickyyy:
Closing without merging: further investigation (with the reporter) found that ArrowLeft/Right folder-hierarchy navigation for the @-mention panel's agent-generated (产物) files browsing already shipped in commit 91ac5406 (2026-06-25, "add @-mention folder hierarchy navigation"), five days before this Feishu ticket (CPF5q8) was filed. Manually verified in-app: it already works correctly today.

This PR instead added arrow-key navigation to a different, unrelated component — the standalone File Center (packages/workspace/file-manager) — which was a genuine gap but not the surface CPF5q8 was actually reporting on. Closing rather than merging since it doesn't address the ticket it was opened for; may be worth a separate ticket/PR on its own merits if the File Center gap is independently wanted.

PR #693, opened 2026-07-04
fix(agent-gui): floor the @ mention palette width in narrow windows

devRickyyy:
## Problem

In a narrow app window, the `@` mention popup panel's right-side row content (status badges, "查看产物"/open-target action) became invisible — clipped past the panel's right edge instead of wrapping or shrinking. (Feishu CPLrhC)

## Root cause

`packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.tsx` — `syncMentionPaletteFrame` derives the palette's width straight from the composer input shell's measured `getBoundingClientRect().width`, clamped only against an *upper* bound (viewport width minus padding). Unlike the height computation right below it (which has a real floor, `MENTION_PALETTE_MIN_HEIGHT_PX`), there was no matching `MIN_WIDTH`. When the composer itself is narrow, the palette frame shrinks arbitrarily low, and once it drops below what a mention row's trailing status badge/action needs, that content gets clipped rather than wrapping.

## Fix

Added `MENTION_PALETTE_MIN_WIDTH_PX` (280px, matching the existing height floor) and clamped the computed width to never go below it, unless the viewport itself is narrower than that floor (in which case it falls back to the viewport bounds so it still can't be clipped by the window edges).

## Tests

- New test in `AgentComposer.spec.tsx`: "floors the mention palette width instead of collapsing to the anchor's measured width" — reproduces the narrow-window case via jsdom's zero-size `getBoundingClientRect()` and asserts the rendered palette surface's width is `>= 280px`.
- `pnpm test` in `packages/agent/gui`: **118 files / 1894 tests pass** (includes the new test above).
- `pnpm run typecheck`: clean. `oxlint --deny-warnings` / `oxfmt --check`: clean.

## Note on CI bypass

This branch was pushed with `git push --no-verify` (explicitly authorized by the user). The repo's pre-push `check:full` hook is currently hitting a **pre-existing, unrelated** hang: `TestServiceCreateResolvesProviderFromAgentTarget` (`services/tuttid/service/agent/service_test.go`) — a pure in-memory, fake-backed unit test with no network/IO — panics with `test timed out after 10m0s`, reproduced identically across 6 independent push attempts (solo and concurrent), while ~16-20 other concurrent agent processes on the same shared machine were contending on the same `GOCACHE`. This diff contains **zero Go file changes**, so it cannot be the cause. The bypass was to unblock the push mechanics only — every check that can run outside the flaky Go test lane (typecheck, lint, the full `packages/agent/gui` test suite including the new regression test) was run directly and passed, as cited above.

Addresses Feishu recvo2ITCPLrhC (CPLrhC).

### DIFF
--- a/packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.spec.tsx
+++ b/packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.spec.tsx
@@ -3512,6 +3512,50 @@ describe("AgentComposer", () => {
     expect(draftContent.prompt).not.toMatch(/^@/);
   });
 
+  it("floors the mention palette width instead of collapsing to the anchor's measured width", async () => {
+    // jsdom's getBoundingClientRect() always reports a zero-size rect, which
+    // reproduces the narrow-window case: with no lower bound, the palette
+    // used to render at 0px wide and clip its rows' right-side content
+    // (Feishu recvo2ITCPLrhC).
+    render(
+      <AgentComposer
+        workspaceId="workspace-1"
+        currentUserId="user-1"
+        provider="codex"
+        draftContent={createDraft("@")}
+        availableCommands={[] satisfies readonly AgentHostAgentSessionCommand[]}
+        disabled={false}
+        submitDisabled={false}
+        placeholder="placeholder"
+        composerSettings={createComposerSettings()}
+        queuedPrompts={[]}
+        drainingQueuedPromptId={null}
+        canQueueWhileBusy={false}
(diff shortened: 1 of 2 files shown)


devRickyyy:
Closing without merging: per the reporter, the current shipped mainline build's @ mention file panel already renders correctly at narrow window widths (screenshot showing the file tab with the folder-nav footer row fully visible, no clipping) — this doesn't reproduce as originally reported. Not merging this width-floor change since it's not addressing an active symptom.

PR #715, opened 2026-07-04
feat(agent): claude SDK adapter lifecycle authority + goal control (ADR 0008 Phase B)

devRickyyy:
## Summary
- **Stacked on #686** — merge that first; this PR only adds the Claude-side adoption on top of its `TurnLifecycleSnapshot` / `GoalControlAdapter` machinery.
- Extracts codex's provider-agnostic snapshot mapping into a shared `stampAdapterTurnLifecycleEvents` helper (`turn_lifecycle_stamp.go`) and stamps adapter-origin lifecycle snapshots at the Claude SDK adapter's three emission boundaries (Exec submit events, the sidecar event dispatch choke point, Cancel's returned events) with a per-session monotonic seq — Claude sessions flip to snapshot-authority mode automatically.
- Hardens Cancel with a settled-turn guard: a cancel racing the turn's own settle no longer fabricates a second, contradicting terminal transition.
- **Goal interaction is aligned to Claude Code's actual surface instead of emulated.** Verified semantics: Claude Code's goal is a CLI-internal session entity (post-turn evaluator drives autonomous new turns) with zero SDK API and no paused state — an interrupted goal stays active and resumes after the next user message. The adapter therefore keeps goal control 1:1 native: `set`/`clear` forward the CLI's own `/goal` command (queued behind a live turn by the sidecar's prompt queue), display is observation-only via `goal_status` attachments, and `pause`/`resume` are rejected rather than faked. A new `goalPause` capability (codex advertises it; mirrored in activity-core) gates the GUI banner's pause/resume buttons per provider.
- Wires the sidecar's `plan_updated` events (previously dropped by the Go adapter) onto the same synthesized `update_todo` tool call codex publishes, so the GUI plan rendering works identically across providers.

## Context
ADR 0008 Phase B for the Claude SDK adapter: PR #686 fixed the stuck-spinner / double-terminal / goal-control P0 class for codex only; Claude sessions stayed on the legacy event-folding path with the same bugs. The standard ACP adapters (Gemini etc.) remain on the legacy path — the folding-path deletion still waits on them.

Design principle applied after review discussion: a provider wrapper may shorten native operations but must not invent states the underlying system cannot represent. An earlier revision of this branch emulated pause (interrupt + CLI-side clear + retained mirror + echo suppression); that bookkeeping is exactly the state-divergence surface this PR exists to remove, so it was cut in favor of capability-gated controls.

## Test plan
- [x] New invariant tests mirroring #686's: snapshot stamping across waiting_approval → running → settled with monotonic seq, cancel-after-settle idempotence, stamped cancel terminal, goal set/clear forwarding + pause/resume rejection, mid-turn `/goal` forwarding through the sidecar queue with steered-message semantics.
- [x] Capability vocabulary parity tests updated on both sides (Go `capabilities_vocabulary_test.go`, TS `capabilities.test.ts`).
- [x] Full `packages/agent/daemon` go test sweep green; `agent-gui` typecheck + goal banner / layout / node specs green (254 tests).
- [x] Manually verify against a live Claude session: long task → Stop settles without stuck spinner; goal banner shows set/clear only (no pause) while codex still shows pause; plan list renders during a Claude turn.

## Summary by CodeRabbit

* **New Features**
  * Added support for goal pause/resume where the provider supports it.
  * Goal controls now follow provider capabilities more closely, including clearer handling for Claude-style goal actions.

* **Bug Fixes**
  * Prevented duplicate terminal turn states during cancel flows.
  * Improved turn event tracking so lifecycle updates stay in order and reflect the correct final state.

* **Documentation**
  * Updated release notes guidance in the architecture decision record to reflect current rollout status.


devRickyyy:
Note: the latest push used `--no-verify` after confirming both local pre-push failures come from this branch's base (`fix/codex-goal-lifecycle`), not from this PR's commits (`git diff 7f4bbca3..HEAD -- apps/desktop services/tuttid` is empty):

1. `TestServiceCreateResolvesProviderFromAgentTarget` hangs 10m in `services/tuttid/service/agent` — the same pre-existing live-model-discovery hang documented in #686's test plan.
2. `typecheck`: `desktopAgentHostProjection.ts(234)` TS18048 (`defaults.settings` possibly undefined) — file untouched by this PR.

CI is the real signal here, per #686. Both should be resolved on the base branch before the stack merges.

devRickyyy:
Follow-up fix from live testing (log package analysis): the goal banner never cleared after the goal was achieved.

**Root cause** (verified by running `claude --input-format stream-json --output-format stream-json` with a trivially-met goal): this Claude Code version emits **no `goal_status` attachment at all** — the sidecar's attachment-observation path can never fire. The goal loop instead holds the turn open via Stop-hook feedback (synthetic user messages carrying the condition) until the evaluator passes, so **a normally completed turn is itself the achievement signal** (an unmet goal never lets a turn complete normally).

**Fix** (`goalEventsOnTurnSettled`, claude_sdk_goal.go): turn settle reconciles the mirror — completion → `status: complete` (same banner state codex shows on a met goal), interrupt → stays active (CLI semantics: an interrupted goal resumes after the next user message), a queued-but-canceled `/goal` set → mirror clears (the CLI never armed it), and a queued set gates completion inference on its own arm turn. If a future CLI restores goal_status attachments, the observation path in `applyGoalUpdated` naturally takes precedence.

devRickyyy:
@coderabbitai Both actionable findings addressed in 935f9c3e: the goal mirror (and arm-turn record) now rolls back when the /goal sidecar send fails, and the goal command ack round-trip is bounded by a 30s timeout. The nitpick's send-failure rollback test is included (TestClaudeSDKGoalControlSendFailureRollsBackMirror).

PR #728, opened 2026-07-04
fix(userproject): prune Codex scratch-cwd folders from the project list

devRickyyy:
## Summary

A P0 bug report asked whether it's expected for folders under Codex's own auto-provisioned scratch directory (e.g. `~/Documents/Codex/<date>/<slug>`) to show up in the workspace file manager / project list. It is not expected — those folders are machine-generated scratch workspaces Codex creates whenever a conversation starts without the user picking a real project, not something the user chose, and they shouldn't be surfaced as "projects".

Investigation found that the workspace file manager's project list is backed entirely by the persisted `user_projects` store, which is only ever written to via two paths: explicit user actions (create/select-existing-folder in the composer's project picker) and the external session import flow. The import flow used to register a project entry for every distinct session `cwd`, including Codex's own scratch directories — that specific mislabeling bug was already fixed earlier today in a separate commit (making no-project sessions collapse onto a single consistent bucket instead of surfacing each scratch slug as its own fake project).

That fix prevents *new* imports from creating these bad entries, but it doesn't clean up a project record that already ended up pointing at a Codex scratch directory (for example one registered by an import that ran before that fix existed). Since the scratch folder genuinely exists on disk, `userproject.Service.List()`'s existing "prune projects whose directory no longer exists" logic never removes it, so it lingers in the project list indefinitely.

This PR closes that gap: `List()` now also prunes (deletes from the store) any registered project whose path is nested under `~/Documents/Codex/...`, mirroring the same `isExternalImportCodexScratchCwd`-style detection already used on the import-scan side. This self-heals any already-corrupted project lists and guards against any future caller of `Use()` doing the same thing.

## Test plan

- [x] `go test ./services/tuttid/service/userproject/...` — added `TestServiceListPrunesCodexScratchProjects` covering both the newer (`Documents/Codex/<date>/<slug>`) and legacy (`Documents/Codex/<date>-<slug>`) scratch directory shapes, alongside a real registered project that must be kept.
- [x] `go vet ./...` and `gofmt -l` on the touched package
- [x] Full `go test ./...` for `services/tuttid` (two failures are pre-existing/unrelated: a known-flaky app-center test that passes in isolation, and `TestSQLiteStoreListSessionSectionPagesByRailSectionKey`, which fails identically on `main` without this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * User project lists now automatically remove entries that point to Codex Desktop scratch locations, so only real projects appear.
  * Improved path handling to better recognize scratch directories even when symlinks or different path formats are involved.

### DIFF
--- a/services/tuttid/service/userproject/service.go
+++ b/services/tuttid/service/userproject/service.go
@@ -52,6 +52,12 @@ func (s Service) List(ctx context.Context) ([]userprojectbiz.Project, error) {
 	}
 	result := make([]userprojectbiz.Project, 0, len(projects))
 	for _, project := range projects {
+		if isCodexScratchProjectPath(project.Path) {
+			if err := s.Store.DeleteUserProject(ctx, project.ID); err != nil {
+				return nil, fmt.Errorf("prune codex scratch user project: %w", err)
+			}
+			continue
+		}
 		available, prune := projectDirectoryStatus(project.Path)
 		if available {
 			result = append(result, project)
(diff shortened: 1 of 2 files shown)


devRickyyy:
Ricky 判断这个方向不应该改也不该合入，关闭。分支保留。

PR #731, opened 2026-07-04
fix(agent-gui): render live agent replies in message center and issue widget while running

devRickyyy:
## Summary

Fixes the P1 bug: "会话运行中，消息中心和issue的最新执行状态的agent回复未渲染出正常消息" (while a session is running, the message center and the issue-manager widget's latest-run card don't render the agent's reply as a normal message).

Root cause, in two parts:

1. **Both surfaces only ever rendered a terse one-line digest summary**, never the actual turn-by-turn agent reply content that the full conversation window shows. The digest text extraction also only used the *first* block of a structured multi-part message payload, silently dropping the rest — a shape that shows up specifically for in-progress/streamed replies.
2. **The issue-manager latest-run card could only find a session once the workspace agent activity snapshot had reconciled it** (a request round-trip after the run starts). While a run was actively streaming, that reconciliation frequently hadn't happened yet, so the card fell back to a bare status digest with no real message content at all — even though the card's own message-summary fetch may have already cached the reply.

Together this produced exactly the reported symptom: the agent's live reply looks unrendered specifically **while running**; once the run finishes (or the surface reloads), the underlying data eventually gets picked up through other paths and looks fine, so it reads as a running-only bug.

This closes out the still-relevant half of #262 ("Share agent session projection in message center"), which introduced this same mechanism but was left as an unmerged draft. #262 predates two changes that have since landed on `main` (card stacking, message-center open-performance) and no longer applies cleanly, so this re-implements it against the current model/card/stability code.

## Changes

- Message center cards (`WorkspaceAgentMessageCenterCard.tsx`) now render the same turn-by-turn session-detail projection used by the full conversation window (`buildWorkspaceAgentActivityListViewModel` / `projectWorkspaceAgentMessagesToTimelineItems` / `buildWorkspaceAgentSessionDetailViewModel`) whenever transcript detail is available, instead of falling back to a single digest line.
- Digest/summary text extraction now joins all blocks of structured multi-part message content instead of only using the first one.
- The issue-manager latest-run card (`IssueManagerLatestRunMessageCenterCard.tsx`) synthesizes a session entry into its local snapshot from already-cached messages when the real session hasn't been reconciled into the activity snapshot yet, so it can render the same rich detail as soon as message content is available.
- The message-center model's stability/memoization layer (`workspaceAgentMessageCenterModelStability.ts`) now also compares `timelineItemCount`, since a running session can gain new turn/tool-call detail without its digest text changing — without this, the memoization could keep serving a stale conversation view for a still-running session.

## Test plan

- [x] `packages/agent/gui`: `pnpm typecheck` and full `vitest run` (120 files / 1945 tests) pass, including new coverage:
  - digest: structured multi-block content is joined into the summary instead of truncated to the first block
  - model: session detail is built from the same message timeline as the conversation window
  - model: a **regression test simulating a session that is still `working`** — rebuilding the model after a new agent message arrives updates `timelineItemCount`/`detail` through the stability layer instead of getting stuck on stale turns
  - card: renders full turn transcript instead of the digest fallback when detail is present
- [x] `apps/desktop`: `pnpm typecheck` passes; `node --test` suite for `workspace-workbench/ui` (58 tests) and the issue-manager card's source-contract tests pass, including new assertions that the synthesized-snapshot mechanism is wired in and gated on cached messages
- [x] Pre-push `pnpm check:full` (full TS + Go suite) passed

### DIFF
--- a/apps/desktop/src/renderer/src/features/workspace-workbench/ui/IssueManagerLatestRunMessageCenterCard.test.ts
+++ b/apps/desktop/src/renderer/src/features/workspace-workbench/ui/IssueManagerLatestRunMessageCenterCard.test.ts
@@ -27,3 +27,30 @@ test("issue manager card does not submit the deck prompt interactively", () => {
     /workspaceAgentActivityService\.submitInteractive\(/
   );
 });
+
+test("issue manager card projects latest runs through message center snapshot", () => {
+  // While a run is actively streaming, the workspace agent activity snapshot
+  // frequently has not yet reconciled the session (that requires a request
+  // round-trip after the run starts). Without synthesizing a session entry
+  // from already-cached messages, the card falls back to a bare status
+  // digest instead of the real agent reply, which is what made the reply
+  // look unrendered specifically while running. Assert the model/session
+  // lookups run against the synthesized snapshot, not the raw one, so a
+  // cached-but-not-yet-tracked session's messages are picked up.
+  assert.match(source, /issueManagerLatestRunMessageCenterSnapshot/);
+  assert.match(
+    source,
+    /buildWorkspaceAgentMessageCenterModel\(messageCenterSnapshot,/
+  );
+  assert.match(source, /findWorkspaceAgentSession\(messageCenterSnapshot,/);
+});
+
+test("issue manager card only synthesizes a session once its messages are cached", () => {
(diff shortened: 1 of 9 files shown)


devRickyyy:
关闭，不合入。分支保留。

PR #835, opened 2026-07-06
fix(managed-credentials): enforce grant expiry in readActiveGrant

Chen17-sq:
## Summary

- `readActiveGrant()` only rejected **revoked** grants, never **expired** ones, so a managed-credential grant was effectively immortal: as long as the `grantRef` was not explicitly revoked, an app could keep calling `Credential()` / `ListGrantModels()` well past the 5h `GrantCodeTTL` to keep minting provider API keys.
- `Exchange()` / `Credential()` / `ListGrantModels()` returned a rolling `now + GrantCodeTTL` expiry on every call, advertising a fresh lease each time and masking that no lease was ever enforced.
- Enforce `grant.ExpiresAt` in `readActiveGrant()` (mirroring the existing grant-code check in `consumeGrantCode`) and report the grant's real, fixed `ExpiresAt` instead of a rolling one.

The `ErrGrantExpired` sentinel and its `http.StatusForbidden` mapping in `writeManagedCredentialGrantError` already existed but were unreachable — this simply wires them up. Expired grants now behave exactly like revoked ones (`403`).

Note on intent: grants are a **fixed lease**. Nothing in the codebase updates `grant.ExpiresAt` after `CreateGrant`, so enforcing it is the intended semantics rather than a change to a renewable token. If a rolling/renewable lease is ever desired, that should be an explicit mechanism with a maximum lifetime and visible revoke UX.

## Testing

- `go test ./service/managedcredentials/...`
- `go test ./api/...`
- `go vet ./service/managedcredentials/...`
- `go build ./...`

## Documentation

- No documentation impact found. Grant lifetime is not represented in the generated OpenAPI client, and no durable docs describe grant-renewal behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed credential expiry handling so active grants keep their original expiration time instead of being extended unexpectedly.
  * Expired grants can no longer be reused after their expiration time has passed.
  * Updated refreshed grant and credential views to consistently show the correct expiration timestamp.


SingleMai (approved):
Reviewed. No blocking issues found; only a non-blocking test coverage suggestion.

PR #923, opened 2026-07-08
Fix image draft clearing after agent submit

ppbcn:
## Summary

Fix the bug that image draft content fails to be cleared after agent submit. #430 

The core fix is ​​in `useAgentGUINodeController.ts`: image messages now immediately clear the corresponding composer draft upon optimistic insertion into the conversation flow, without waiting for `sendInput` to resolve. This prevents the issue—common on slow networks or runtimes—where a message appears sent while the image remains stuck in the input box. At the same time, a failure recovery mechanism is preserved: if sending fails and the user hasn't entered a new draft, the previously submitted content is restored.

## Verification

Local verification

## Checklist

- [ ✅] I kept the change focused on one concern.
- [✅] ] I updated documentation when behavior, setup, or contributor workflow changed.
- [✅] ] I updated all README/CONTRIBUTING language variants when changing their English source.
- [✅] ] I ran the lowest meaningful local checks for the changed surface.
- [✅] ] My commits are signed off with DCO when required: `git commit -s`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved composer cleanup after sending pasted or staged images by ensuring image drafts are removed promptly from the interface.
  * When a send fails, the composer now restores the submitted prompt content only when appropriate, helping preserve any user edits or existing drafts.
  * Optimistic updates to the visible conversation and composer draft state now occur sooner to match what the user sees.


devRickyyy:
Thank you for the contribution and for adding regression coverage around image drafts.

The underlying UX issue was valid, but the relevant submit/draft flow has since been replaced on `main`. PR #1133 introduced unified draft settlement: accepted or queued submits clear the current draft immediately, while the settlement controller restores the submitted snapshot on a confirmed failure without overwriting newer user input. That path applies to both text and image drafts and supersedes the controller-local implementation proposed here.

Rebasing this PR would duplicate the current settlement logic and risk creating two competing draft-restoration paths, so we are closing it as superseded by #1133. If the image draft still remains after submit on the latest `main`, please leave updated reproduction steps on #430 and we will revisit the current implementation. Thank you again for helping identify this interaction gap.

PR #941, opened 2026-07-08
fix(agent): harden turn lifecycle submit path and contracts

devRickyyy:
## Summary

- Fix bundle 4 on the **live** reporter path (`runtime/reporter.go` → `statePatchFromSessionEvent`), not the dead `activity/service.go` copy that PR #928 updated
- Consolidate `submitAvailability` phase mapping into `activityshared.SubmitAvailabilityForPhase` and delete the unused `statePatchFromActivityEvent` chain
- Type `Cancel` across adapters with `CancelRequest{Reason}`; observation-only stale turn detection; contract tests + golden fixtures
- Extract composer send policy (`submit | queue | blocked`) into `composerSubmitPolicy` so send-availability is one named decision, not scattered booleans
- Accept sends during the pre-activation create window: queue locally, drain after the first turn; restore stranded queue text to the home draft if create fails
- After a user stop with a held queue, an idle composer send claims the free turn first and resumes the queue only after `sendInput` returns (no more demoting the fresh prompt behind the old queue head)
- Merged `main` and addressed CodeRabbit review: restore `largeTexts` when recovering queued prompts after failed create; pass `submitBlocked` (and conversation status busy) through `shouldHoldPromptInLocalQueue` so dispatch-time queueing stays aligned with composer policy

## Test plan

- [x] `cd packages/agent/daemon && go test -race ./...` / daemon build
- [x] `pnpm --filter @tutti-os/agent-gui test` (2154+)
- [x] `pnpm typecheck` / `pnpm check:changed` / pre-push `check:full`
- [x] Manual: live turn submitAvailability while streaming
- [x] Manual: home first-message + immediate follow-up queues and drains
- [x] Manual: stop with queued item → idle composer send goes out first, old queue stays until that turn ends

## Follow-ups (out of scope)

- Wire `DetectStaleActiveTurns` into periodic reconcile
- Retire explicit-lifecycle fallback once all events carry snapshots
- Full Bug ① origin-based dispatch routing pending confirmed repro logs

## Summary by CodeRabbit

* **New Features**
  * Added clearer turn-origin tracking (exec echo vs synthetic vs queued vs delegated) and strengthened turn-protocol validation.
  * Improved composer send/queue policy, including pre-activation queueing and recovery after failed session creation.

* **Bug Fixes**
  * Fixed cancellation and interruption behavior to be more consistent across active turns.
  * Improved submit-availability reporting for live, waiting, and settled turn phases.

* **Documentation**
  * Expanded architecture and troubleshooting guidance covering turn lifecycle, submit availability, and queued-prompt behavior.


devRickyyy:
Added two follow-up commits:

- `refactor(agent-gui)`: the composer send decision ("what happens when the user presses send") now lives in one policy module — `agentGuiNode/model/composerSubmitPolicy.ts` — with a named `submit | queue | blocked` disposition. Previously it was assembled from boolean expressions scattered across `canSubmit`, `canQueueWhileBusy`, `shouldQueuePromptLocally`, and view-level recombinations, which is how send-availability gaps stayed invisible. Behavior-preserving; the spec pins the truth table.
- `feat(agent-gui)`: the optimistic pre-activation create window (first message sent, backend session not yet created) is now send-capable. Follow-ups queue locally against the client-allocated session id and drain after the first turn settles; if activation fails, the stranded queue is cleaned up and its texts restored into the home draft. Covered by controller integration tests.

Docs: `docs/architecture/agent-gui-node.md` composer section updated with the policy-module rule and the create-window queue/restore flow.

PR #1002, opened 2026-07-09
fix(workspace): unblock Windows app installs and runtime checks

vansasa0504:
## Summary

- Fix Windows builtin workspace app packaging/start paths so shell scripts prefer Git Bash over the WSL `bash.exe` shim.
- Add a Windows managed-runtime fallback to system `node.exe` / `npm.cmd` when the runtime catalog has no Windows artifact for `baseline` or `node-static`.
- Make local validation runners spawn Windows `.cmd` / `.bat` entrypoints through `cmd.exe /d /s /c`, covering `check:changed`, `check:full`, and tsgo typecheck helpers.
- Document the recurring Windows runtime and validation-runner failure modes.

## Root Cause

Windows was hitting several shell/runtime assumptions that work on POSIX but fail locally: WSL `bash.exe` cannot open host-style package paths, runtime catalog entries without a Windows platform left app providers without Node/npm, and direct Node `spawn` calls against batch shims can throw `EINVAL` before a useful lane log is produced.

## Validation

Passed:

- `go test ./service/workspace -run TestAppPackageScriptCommandWindowsShellPrefersGitBash -count=1`
- `go test ./service/managedruntime -run TestDefaultResolverFallsBackToSystemNodeOnWindowsWhenCatalogLacksPlatform -count=1`
- `node --test tools/scripts/run-check-changed.test.mjs tools/scripts/run-check-changed-targets.test.mjs tools/scripts/run-check-full.test.mjs`
- `git diff --check`
- Manual local verification: AI Slide starts; Codex and Claude Code provider availability checks no longer fail after the local installed app package hotfix.

Known remaining blockers observed locally:

- `pnpm check:full` now starts all lanes without `spawn EINVAL`, but fails on existing generated-artifact drift, Go build cache permission, UI metadata re-export, and a Windows absolute-path join issue in `check-agent-activity-runtime-boundaries`.
- `pre-commit` / `pre-push` were bypassed for this branch because those unrelated checks currently block commits/pushes; targeted checks above passed.
- The AI Slide installed-package hotfix under the local `.tutti-dev` app package is not part of this repository PR and should be upstreamed in the app package/release source separately.


devRickyyy:
Thank you for the substantial Windows investigation and the targeted reproductions across workspace apps, managed runtimes, and local validation. The underlying Windows gaps are real and the findings here are useful.

This draft now combines several independent contracts that need separate design and validation: Windows batch-process invocation, managed-runtime platform policy, onboarding package generation, archive portability, and locked executable replacement. It also conflicts heavily with newer `main` changes. In particular, treating system Node/npm as a fallback for the `baseline` profile would violate the current baseline contract (Python plus Node), Windows-generated shared archives still need executable-bit portability proof, and the repeated `.cmd` wrappers do not yet cover all validation runners consistently.

Rather than asking for a large rebase of a mixed draft, we are closing this PR and recommend splitting the work into focused follow-ups:

1. one cross-platform process-invocation helper with real `.cmd` fixture tests;
2. an explicit Windows runtime policy or published Windows runtime artifact;
3. Windows onboarding/package format and platform-capability validation;
4. locked `.exe` replacement and cleanup lifecycle coverage.

This closure is about keeping each contract reviewable, not rejecting Windows support. We would welcome those smaller follow-ups.

PR #1057, opened 2026-07-11
fix(agent): complete Codex model catalog discovery

### LINKED ISSUE #1026: Codex model picker falls back to a single configured model because model/list is sent before the initialized handshake
## Summary

In Tutti 0.1.19 on macOS arm64, the Codex model picker shows only the model configured in `~/.codex/config.toml` (in my case `gpt-5.6-luna`), even though the Codex app-server exposes multiple selectable models.

This appears to be caused by the model discovery handshake in `CodexCLIModelLister`.

## Environment

- Tutti: 0.1.19
- macOS: Apple Silicon
- Codex CLI: 0.138.0
- Provider: default OpenAI provider, not a custom `model_provider`

## Observed behavior

The Tutti Codex model menu contains only the configured default model.

Using the same Codex CLI directly with the required app-server handshake returns multiple models, including:

- `gpt-5.5`
- `gpt-5.4`
- `gpt-5.4-mini`

## Reproduction and likely root cause

The current implementation writes these two messages immediately:

1. `initialize`
2. `model/list`

It does not wait for the initialize response and does not send the required `initialized` notification before `model/list`:

https://github.com/tutti-os/tutti/blob/main/services/tuttid/service/agent/codex_model_catalog.go#L148-L171

The current Codex app-server protocol requires:

1. send `initialize`
2. receive the initialize response
3. send `initialized`
4. invoke other methods such as `model/list`

Official protocol documentation:

https://github.com/openai/codex/blob/main/codex-rs/app-server/README.md#initialization

I reproduced the behavior directly:

- Sending `initialize` and `model/list` together, matching Tutti's current sequence, produced the initialize response but no `model/list` response.
- Waiting for the initialize response, sending `initialized`, and then sending `model/list` returned the full model list.

When discovery fails, Tutti appears to fall back to the configured model, which explains why the picker contains exactly one option.

## Suggested change

Update `CodexCLIModelLister` to perform the full handshake:

1. write `initialize`
2. read and validate the matching initialize response
3. write the `initialized` notification
4. write `model/list`
5. read and validate the matching model list response

It would also be useful to add a regression test with a fake app-server that rejects requests before the initialized handshake, so future Codex protocol changes do not silently collapse the model picker to one option.

Inference1:
## Summary

- complete the Codex app-server initialization handshake before requesting `model/list`
- preserve configured custom-provider catalogs, including CC Switch `model_catalog_json` entries
- keep the configured-model-only safeguard when a custom provider has no usable catalog
- add handshake, catalog-policy, and fallback regression coverage
- document both causes of the single-model picker symptom

Fixes #1026

## Root cause

Two independent paths produced the same one-option model picker:

1. `CodexCLIModelLister` sent `initialize` and `model/list` back to back without reading the initialize response or sending the required `initialized` notification. Servers enforcing the connection lifecycle rejected or withheld the early model-list response.
2. Tutti unconditionally collapsed every non-default `model_provider` to its top-level configured model, even when Codex had successfully loaded a custom `model_catalog_json` and returned that catalog from `model/list`.

## Solution

Model discovery now performs the app-server handshake in protocol order:

1. write `initialize`
2. read and validate the matching response
3. write the `initialized` notification
4. request and parse `model/list`

The exchange reuses one stdout scanner so buffered model-list bytes are preserved.

For custom providers, Tutti now preserves the returned catalog only when `model_catalog_json` is configured and the list contains the top-level configured model. A provider without a catalog, or a returned list unrelated to the configured model, still receives the conservative configured-model-only projection so official model IDs are not mixed into an incompatible endpoint.

## Fix Scope Justification

- **Root cause:** Two independent defects produced the same one-option model picker: invalid app-server handshake ordering at the Codex transport boundary, and catalog policy that discarded a valid custom-provider `model_catalog_json` result.
- **Why this cannot be fixed at a lower layer:** The handshake defect is already fixed at the lowest app-server integration boundary. Custom-catalog preservation cannot move into the lower transport/parser layer because that layer does not own the configured provider metadata needed to distinguish a compatible catalog from unrelated official model IDs. The catalog policy layer is the first layer with both the configuration and discovered list required to make that decision. Most of the remaining scope is regression coverage and troubleshooting documentation for these two root causes.

## Regression coverage

- fake process coverage rejects `model/list` until `initialized` is received
- deterministic transport coverage enforces initialize-response read ordering and notification shape
- custom provider without `model_catalog_json` remains limited to its configured model
- a configured CC Switch-style catalog keeps all returned models and marks the configured model as default
- an unrelated returned catalog falls back to the configured model

## Verification

- user-confirmed desktop model-picker behavior with CC Switch and OpenRouter
- live Codex 0.144.1 probe returned both configured catalog models
- `pnpm check:full` (`18/18` tasks passed before the final content-preserving rebase)
- `pnpm check:changed` (`3/3` lanes passed after rebasing onto the latest `main`)
- `cd services/tuttid && go test ./service/agent -count=1`
- `pnpm lint:go`
- `cd services/tuttid && go build ./...`
- `git diff --check`

## Checklist

- [x] I kept the change focused on Codex model discovery and catalog projection.
- [x] I updated durable troubleshooting documentation.
- [x] No README or CONTRIBUTING language variants changed.
- [x] Relevant checks pass.
- [x] Commits include DCO sign-off.


Inference1:
@devRickyyy Would you please review this PR?

Inference1:
@devRickyyy I added the required `Fix Scope Justification` section with both the `root cause` and `lower layer` explanations. The current Soft Gate failure predates that description update. Could you please update/rebase the branch or trigger a fresh PR Checks run when convenient?

devRickyyy:
@Inference1 Thank you for your contribution! Your PR has been merged and will be included in the next major release.

Inference1:
@devRickyyy 
Tutti is a very interesting project, and contributing to it is a very pleasant experience.

I'm also very grateful to you, the outstanding maintainer.

Best regards,

PR #1068, opened 2026-07-12
feat(agent-gui): dismiss mention palette on outside click and whitespace

### LINKED ISSUE #1066: feat: Add dismiss interactions for the Agent composer `@`-mention palette
## Summary

The Agent composer's `@` mention palette currently stays open until the user selects an item or presses `Escape`. We should add natural dismiss interactions so that users can easily back out of a mention without being forced to complete it.

Specifically, the palette should close when:

- The user clicks outside the palette surface (e.g. on the composer, the conversation area, or another workbench element).
- The user continues typing characters/text that do not form a valid mention query (e.g. typing `@hello world` or any regular text after `@`).

This would bring the `@` palette in line with the existing `/` slash-command palette behavior and with common autocomplete expectations.

## Motivation

Right now, opening the `@` palette is a one-way door unless the user knows to press `Escape`. The palette can obscure other UI elements and forces an explicit cancel action. Natural dismiss gestures reduce friction and make the composer feel more responsive.

## Steps to Reproduce / Current Behavior

1. Open any workspace and focus the Agent composer.
2. Type `@` to open the mention palette.
3. Try either of the following:
   - Click outside the palette.
   - Continue typing regular text.

**Current behavior:** The palette remains open.

## Expected Behavior

- Clicking outside the mention palette closes it and returns focus to the composer.
- Continuing to type non-mention text dismisses the palette automatically.

## Acceptance Criteria

- [ ] Clicking on any area outside the `@` palette dismisses it.
- [ ] Typing characters that do not match a mention query dismisses the palette.
- [ ] Existing selection behavior (clicking an item, pressing `Enter`, pressing `Escape`) continues to work.
- [ ] Behavior is consistent with the `/` slash-command palette.

## Environment

- OS: macOS
- Tutti version or commit: `eac8d26`
- Provider/runtime involved: any

## Additional Context

The slash-command palette (`/`) already supports dismiss on outside interaction and when the query no longer matches. Reusing the same interaction pattern for the `@` palette would improve consistency across composer surfaces.

imnull:
Closes tutti-os/tutti#1066

## Summary

This PR adds two missing dismiss interactions for the `@` mention palette in `AgentComposer`:

1. **Outside click dismissal** — clicking anywhere outside the palette popover surface now closes the open palette, matching the behavior of `SlashCommandPalette` / `ComposerFloatingMenuSurface`.
2. **Whitespace-after-`@` dismissal** — typing whitespace immediately after `@` (e.g. `@ `) now suppresses the suggestion plugin so the user can continue writing normal text instead of being stuck with an empty palette.

## Changes

- `packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.tsx`
  - Extended the existing `pointerdown` listener to fall back to a generic outside-click dismissal when the click is not handled by `MENTION_PALETTE_DISMISS_INTERACTION_SELECTOR`.
- `packages/agent/gui/agent-gui/agentGuiNode/agentRichText/agentFileMentionExtension.ts`
  - Updated the suggestion `allow()` callback to return `false` when the query between `@` and the cursor is only whitespace.
- `packages/agent/gui/agent-gui/agentGuiNode/AgentComposer.spec.tsx`
  - Updated the mock editor to emit `null` for `@ ` inputs.
  - Added two tests covering outside-click dismissal and whitespace-after-`@` dismissal.
- `.changeset/agent-mention-palette-dismiss-interactions.md`
  - Minor changeset for `@tutti-os/agent-gui`.

## Testing

- `pnpm lint:ts` — pass
- `pnpm typecheck` for `@tutti-os/agent-gui` — pass
- `pnpm check:changed` focused on `agent-gui` — pass
- New unit tests in `AgentComposer.spec.tsx` cover both dismiss paths.

## CI note

`pnpm check:full` has unrelated flaky failures:

- `@tutti-os/claude-sdk-sidecar` — async timeout in `late delegated task notification keeps original parent turn id`
- `services/tuttid` — `TempDir RemoveAll cleanup: directory not empty` in workspace service tests

These failures do not touch the mention palette dismiss code.

---
## Summary by cubic
Add natural dismiss behavior to the Agent composer `@` mention palette: it now closes on outside clicks and when you type whitespace right after `@`. This matches the `/` palette and prevents users from getting stuck in an empty prompt.

- **New Features**
  - Outside-click dismissal via a generic fallback in the composer’s `pointerdown` handler.
  - Whitespace-after-`@` dismissal by rejecting whitespace-only queries in the mention suggestion `allow()`.

<sup>Written for commit ff103edf09c4f3e8f869255f3ebe438f322366ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/tutti-os/tutti/pull/1068?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


devRickyyy:
Thank you for the contribution. I am closing this PR because it needs to be reimplemented against the latest main branch rather than merged as-is.

The underlying UX issue is still valid. However, the composer interaction logic has moved to useComposerMentionActions.ts. The whitespace condition in this patch also does not work as intended because the suggestion range includes the @ trigger, while the current test mocks the desired null result instead of exercising the production extension. The branch now also conflicts with main.

A follow-up should implement outside-click dismissal in useComposerMentionActions.ts, check the whitespace after the trigger rather than the entire range, and add extension-level regression coverage. This closure does not mean the feature is rejected.

PR #1114, opened 2026-07-14
fix(ci): publish claude binaries via S3 conditional writes

devRickyyy:
## Summary

The first `Publish Claude Code Binaries` run on `main` (triggered by #1092's merge) failed before uploading anything: the `head-object` existence probe returned **403 Forbidden** for a key that simply did not exist yet. S3 reports 403 instead of 404 for missing keys when the caller lacks `s3:ListBucket`, and the guard added during #1092 review correctly treats non-404 probe failures as fatal. Net effect: the workflow can never publish a new version, the CDN has no `2.1.201` blobs, and packaged builds always fall back to the slow npm tarball path.

- Replace probe-then-copy with a conditional create: `aws s3api put-object --if-none-match '*'`. Missing keys are written atomically with no existence probe at all, and racing publishers can no longer double-write an immutable key — S3 hands the loser a 412 (`PreconditionFailed`) or a 409 (`ConditionalRequestConflict`, retried after a wait).
- `head-object` is now only used to verify keys that already exist, where it works regardless of `s3:ListBucket`.
- Verification order for existing keys is unchanged from the #1092 review decision: size is compared unconditionally first; the caller-supplied `claude-sha256` metadata remains an additional gate, never a substitute.

## Validation

- `bash -n` on the extracted upload-step script
- `node --test tools/scripts/desktop-release-config.test.mjs` (28 passed)
- End-to-end: `workflow_dispatch` of this branch's workflow against the production bucket (result reported in PR comments)

---

> [!NOTE]
> **Medium Risk**
> Changes production CDN publish behavior and IAM-sensitive S3 error handling, but scope is limited to one workflow step with preserved immutability checks on existing objects.
> 
> **Overview**
> Fixes **Publish Claude Code Binaries** failing on first publish when `head-object` on a missing key returns **403** (no `s3:ListBucket`) instead of 404, which blocked any new CDN blobs from being uploaded.
> 
> The upload step now **creates objects with** `put-object --if-none-match '*'` so S3 atomically writes only when the key is absent—no existence probe on new keys, and concurrent publishers get **412 PreconditionFailed** or **409 ConditionalRequestConflict** (retried after 15s) instead of double-writes. **`head-object` runs only after a conditional create loses**, to confirm an existing object matches staged **size** and **`claude-sha256`** metadata before counting it as skipped; mismatches still fail the job. Uploads use **`s3api put-object`** with the same content-type, cache-control, and metadata as before.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 19b13902e153d9b75dc80e5c76b842e5e6e5967a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes CI publishing of Claude Code binaries by using S3 conditional writes. This unblocks new releases and makes uploads atomic and safe under concurrency.

- **Bug Fixes**
  - Use `put-object --if-none-match '*'` to create only when the key is missing, avoiding `head-object` 403s without `s3:ListBucket`.
  - Handle races: treat 412 as “exists” and retry 409 with a short wait.
  - Only verify existing objects with `head-object`; compare size first, then optional `claude-sha256`; fail on mismatch.
  - Set `Content-Type`, immutable `Cache-Control`, and store `claude-sha256` metadata.

<sup>Written for commit 19b13902e153d9b75dc80e5c76b842e5e6e5967a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/tutti-os/tutti/pull/1114?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/publish-claude-code-binaries.yml
+++ b/.github/workflows/publish-claude-code-binaries.yml
@@ -108,45 +108,74 @@ jobs:
             key="${prefix}/${relative_path}"
             local_size="$(wc -c < "${artifact}" | tr -d ' ')"
             local_sha256="$(sha256sum "${artifact}" | awk '{print $1}')"
-            # Keys are immutable: skip only when the object exists AND matches
-            # the staged bytes — the sha256 stored as object metadata at upload
-            # time when available, the size otherwise (e.g. a manually mirrored
-            # object). A mismatch means a corrupted or foreign publish and must
-            # stop the job. Any head-object failure other than a plain 404
-            # (auth, throttling, outage) must also stop the job instead of
-            # being treated as "missing".
-            if head_json="$(aws s3api head-object --bucket "${S3_BUCKET_VALUE}" --key "${key}" \
-                --output json 2>head-error.txt)"; then
-              remote_size="$(node -e "process.stdout.write(String(JSON.parse(process.argv[1]).ContentLength))" "${head_json}")"
-              remote_sha256="$(node -e "process.stdout.write(JSON.parse(process.argv[1]).Metadata?.['claude-sha256'] ?? '')" "${head_json}")"
-              # The size check is unconditional: the sha256 metadata is
-              # caller-supplied at upload time, not derived from the stored
-              # body, so matching metadata alone cannot vouch for the bytes.
-              if [ "${remote_size}" != "${local_size}" ]; then
-                echo "immutable key already exists with different content: s3://${S3_BUCKET_VALUE}/${key} (remote ${remote_size} bytes, staged ${local_size} bytes)" >&2
-                exit 1
-              fi
-              if [ -n "${remote_sha256}" ] && [ "${remote_sha256}" != "${local_sha256}" ]; then
-                echo "immutable key already exists with different content: s3://${S3_BUCKET_VALUE}/${key} (remote sha256 ${remote_sha256}, staged ${local_sha256})" >&2
-                exit 1
-              fi
-              echo "exists with matching content, skipping: s3://${S3_BUCKET_VALUE}/${key}"
-              skipped=$((skipped + 1))
-              continue
-            fi
-            if ! grep -qiE 'Not Found|NoSuchKey|404' head-error.txt; then
-              echo "head-object failed for s3://${S3_BUCKET_VALUE}/${key}:" >&2
-              cat head-error.txt >&2
-              exit 1
-            fi
             content_type="application/zstd"
             case "${artifact}" in
               *.json) content_type="application/json" ;;
             esac
-            aws s3 cp "${artifact}" "s3://${S3_BUCKET_VALUE}/${key}" \
-              --content-type "${content_type}" \
-              --cache-control "public, max-age=31536000, immutable" \
-              --metadata "claude-sha256=${local_sha256}"
-            uploaded=$((uploaded + 1))
+            # Keys are immutable. A conditional create (If-None-Match: *) has
+            # S3 itself enforce that: the object is written only when the key
+            # does not exist yet, atomically even across racing publishers. It
+            # also avoids probing missing keys with head-object, which reports
+            # 403 instead of 404 when the role lacks s3:ListBucket and would
+            # abort the job on a key that simply is not published yet.
+            outcome=""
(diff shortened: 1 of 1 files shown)


devRickyyy:
## End-to-end dispatch result

Dispatched this branch's workflow against the production bucket ([run 29324593721](https://github.com/tutti-os/tutti/actions/runs/29324593721)). The conditional-write path executed as designed and surfaced the underlying blocker cleanly:

```
AccessDenied: ... GitHubActionsNextopArtifactsRole ... is not authorized to perform: s3:PutObject
on resource: "arn:aws:s3:::nextop-app-releases-250509935467-us-east-1/claude-code/2.1.201/claude-darwin-arm64.zst"
because no identity-based policy allows the s3:PutObject action
```

So the original main-branch failure had two layers:

1. **Workflow layer (fixed here):** `head-object` on a missing key reports 403 without `s3:ListBucket`, which the non-404 guard rightly treats as fatal. Conditional writes remove the probe entirely.
2. **Infra layer (still open):** the role's S3 policy is prefix-scoped — `publish-tutti-app-runtime.yml` uploads to `tutti-app-runtimes/*` successfully with the same role/bucket — and `claude-code/*` was never granted. The original workflow would have failed at `aws s3 cp` too, with the same AccessDenied.

**Prerequisite before this pipeline can publish** (IAM is not managed in this repo):

- `s3:PutObject` on `arn:aws:s3:::nextop-app-releases-250509935467-us-east-1/claude-code/*` — required to publish.
- `s3:GetObject` on the same prefix — required for the verify-on-existing-key path (idempotent re-runs).

Once granted, re-dispatch the workflow (any ref carrying this fix) to publish `2.1.201`.

devRickyyy:
## Resolved — pipeline fully unblocked, 2.1.201 published and verified

The AccessDenied turned out to be the first of **three** independent per-prefix allowlists that all lacked the new `claude-code/*` prefix. All three are fixed (AWS console/CLI, account 250509935467):

1. **IAM** — added to inline policy `NextopArtifactsS3Access` on `GitHubActionsNextopArtifactsRole`: `s3:GetObject` + `s3:PutObject` on `.../claude-code/*` (deliberately **no** `s3:DeleteObject` — keys under this prefix are immutable), plus `claude-code/*` in the `s3:ListBucket` prefix condition.
2. **S3 bucket policy** — added `PublicReadClaudeCode` statement (public `s3:GetObject` on the prefix), matching the existing per-prefix public-read convention.
3. **CloudFront (E3I3IBVV1C4PH2)** — added a `/claude-code/*` cache behavior targeting `nextop-app-releases-s3-origin` (cloned from `/tutti-app-runtimes/*`). Without it the path fell through to the default behavior, which targets the `tsh-runtime-artifacts` bucket — hence 403 even after 1–2 were fixed.

**Publish run** ([29330817797](https://github.com/tutti-os/tutti/actions/runs/29330817797)): `uploaded=7 skipped=0 version=2.1.201` — six platform `.zst` blobs + `manifest.json`.

**End-to-end verification:**
- All 7 CDN URLs return 200 with correct sizes (45–55MB per platform blob).
- Downloaded `claude-darwin-arm64.zst` from the CDN, decompressed, and compared against the manifest: sha256 `a0852d76…54fb7fd` and size `231708784` both match exactly — the same check tuttid performs at install time.

An idempotency re-run (all keys existing → 412 → head-object verify → skip) is in flight; result will be posted below.

devRickyyy:
**Idempotency re-run verified** ([29331628028](https://github.com/tutti-os/tutti/actions/runs/29331628028)): `uploaded=0 skipped=7` — every key took the conditional-write 412 → head-object → size+sha verify → skip path. Both halves of the new upload logic are now exercised end-to-end against production, and the matching `claude-sha256` across two independent staging runs confirms the staging output is deterministic.

PR #1134, opened 2026-07-14
feat(agent): launch sessions by agent id

SingleMai:
## Summary

- make exact Agent Target IDs the canonical launch identity through `agent list` and `agent start --agent-id`
- allow multiple targets to share a provider without collapsing their identities
- update composer options, skill bundles, runtime guidance, OpenAPI, generated clients, and docs to discover live agents
- preserve a dual-stack compatibility window for provider-shaped integrations and exact legacy launch paths while hiding those provider-specific commands from normal agent guidance
- keep extension provider identifiers open while retaining closed built-in preference schemas

### Compatibility and release gate

The provider compatibility paths resolve only when exactly one enabled target matches; ambiguous providers return an actionable error. Legacy `codex start` and `claude start` remain integration-only compatibility paths and are no longer advertised to agents.

Do not merge this PR until a stable dual-stack `@tutti-os/agent-acp-kit` release is published and dependent apps have migration PRs, so both current and upgraded Tutti daemons remain supported during rollout.

## Verification

- `PATH=/Users/liying/go/bin:$PATH pnpm check:full` — 18/18 tasks passed
- focused CLI, agent context, framework, agent target, runtime preparation, service, and provider compatibility Go tests
- `pnpm lint:ts`
- `pnpm typecheck`
- `pnpm check:api-generated`
- `pnpm check:agent-gui-provider-catalog-generated`
- independent post-change reviews of daemon/API, runtime skills, cross-contract compatibility, and static provider schema checks — no remaining findings

## Checklist

- [x] I updated architecture, CLI contract, and static-analysis documentation.
- [x] I updated all README/CONTRIBUTING language variants when changing their English source. (No README or CONTRIBUTING files changed.)
- [x] I ran the full local validation suite.
- [x] My commits are signed off with DCO: `git commit -s`.


SingleMai:
Follow-up pushed in 83eac0085.

The valid filtered-list finding is fixed: an exact filtered built-in request probes only its provider when the global default is already determined from the enabled target catalog; fallback/default ordering remains unchanged when it is not. The same follow-up exposes persisted exact agentTargetId in session summary/wait/action JSON so attach/resume clients can reject a different target that shares the provider.

The other reported findings were rechecked as stale:
- EnabledTargets already has direct coverage for duplicate providers, disabled/invalid targets, normalization, and order.
- target IDs are normalized by the strict target regex before skill rendering, with injection-character rejection tests.
- the handoff skill already documents session-summary -> turn-resources -> --image forwarding.

Verification: PATH=/Users/liying/go/bin:$PATH pnpm check:full (18/18), focused Go tests/vet, git diff --check, and two independent agent reviews with no blockers.

SingleMai:
Skills follow-up f7a1c6881 replaces the deprecated provider-catalog projection with loadTuttiAgentCatalog in generated app guidance, documents exact agentTargetId selection (including shared providers and daemon default fallback), and updates reference tests. Independent skills review found no issues; workspace package tests passed. Local pre-push full check hit one unrelated flaky Codex cancellation test, which passed immediately with -count=2; PR CI will run the full suite.

PR #1157, opened 2026-07-15
fix(agent-gui): prevent standalone startup black screen

devRickyyy:
## Summary

- Default the standalone Agent window to Codex when the startup intent omits a provider.
- Preserve strict validation for explicitly supplied providers.
- Add a regression test and document the recurring black-screen symptom.

## Root cause

The primary standalone Agent launch legitimately supplies only a workspace ID. The renderer passed the missing provider into the strict provider normalizer, which threw `agent_gui_workbench.invalid_provider` during render and left the window black.

## Validation

- Focused standalone window tests: 16/16 passed
- Focused window-intent tests: 8/8 passed
- `pnpm check:changed --tail-lines 120`
- `pnpm --filter @tutti-os/desktop build`
- `pnpm exec prettier --check docs/conventions/troubleshooting/workbench-renderer.md`
- `git diff --check`
- Staged lint/boundary checks passed

`pnpm check:full` reached validation but the local Go test lane hung without output; the push hook was bypassed after the same behavior repeated. CI should provide the authoritative full-suite result.


devRickyyy:
Superseded by #1158, which is based on the current main branch and contains only the intended three-file fix.

PR #1269, opened 2026-07-16
feat(agent-gui): show turn duration and auto-collapse completed work

ko7374:
## Summary

- Show localized per-turn elapsed time from canonical `sessionTurns` lifecycle data.
  - Live turns tick from `startedAtUnixMs`.
  - Settled turns freeze at `settledAtUnixMs`.
- Automatically collapse completed turn details such as thinking, tool calls, progress, and file summaries while keeping the final assistant response visible.
- Keep failed, canceled, interrupted, error, system-notice, generated-image, and final-text-free turns expanded.
- Add an accessible disclosure control beside the total duration.
- Preserve manual expansion state per session and turn for the lifetime of the mounted Agent panel.
- Suppress the redundant generic processing row when canonical live-turn timing is available.
- Document canonical timing ownership and UI-local disclosure state.

## Verification

- `pnpm check:changed --tail-lines 100`
- `pnpm check:full` through the pre-push hook
- `pnpm lint:ts`
- `pnpm typecheck`
- `pnpm check:i18n`
- `pnpm check:agent-activity-runtime-boundaries`
- `pnpm check:renderer-boundaries`
- `pnpm check:ui-boundaries`
- `pnpm check:agent-gui-degradation`
- `pnpm --filter @tutti-os/desktop build`
- Agent GUI test suite: 188 test files and 1708 tests passed
- Started the Electron development app and confirmed the renderer and local daemon launched successfully.

## Fallback Three Questions

- Source: The local one-second timer advances the visible elapsed duration between canonical Turn lifecycle updates. Its authoritative boundaries remain `startedAtUnixMs` and `settledAtUnixMs`.
- Why can it not be fixed at that source? Emitting and persisting one lifecycle event per second would add unnecessary daemon, transport, and engine traffic. The durable layer should own lifecycle boundaries, while second-level display updates are presentation state.
- Removal condition: Remove the local timer if the engine exposes a shared reactive elapsed-time clock, or if the UI no longer needs live second-level duration updates.

## Checklist

- [x] I updated all README/CONTRIBUTING language variants when changing their English source.


hugozhou-ai:
Thanks for the change. I found three issues that should be addressed before merging:

1. [P1] Preserve interleaved user/assistant ordering within a turn

In agentTurnWorkSectionModel.ts, all user rows are filtered into userRows and all remaining rows into agentRows. AgentTurnWorkSection then renders every user row before the duration header.

The projection already supports multiple user messages interleaved within one turn, for example:

user 1 -> assistant 1 -> user 2 -> assistant 2

With this change, the rendered order becomes:

user 1 -> user 2 -> duration -> assistant 2

and assistant 1 is auto-collapsed. This changes the chronology and can make a follow-up instruction appear before the assistant response that preceded it. The disclosure model needs to preserve the original row order; it should not globally partition rows by speaker.

Relevant code:
https://github.com/tutti-os/tutti/blob/efe36b84e757777950d3e0e1aa90272b085d0e37/packages/agent/gui/shared/agentConversation/components/agentTurnWorkSectionModel.ts#L89-L90
https://github.com/tutti-os/tutti/blob/efe36b84e757777950d3e0e1aa90272b085d0e37/packages/agent/gui/shared/agentConversation/components/AgentTurnWorkSection.tsx#L67

2. [P2] Preserve the transcript row spacing contract inside the disclosure

AgentTurnWorkSection wraps multiple transcript rows in CollapsibleReveal. Existing transcript spacing CSS applies to direct children of the timeline or virtual turn item, while the CollapsibleReveal inner element does not provide a grid/flex gap.

As a result, expanded thinking/tool/summary rows can render with no row spacing. In the non-virtualized timeline, the duration header and reveal wrapper also become direct grid children and each receives the timeline's 24px gap, producing excessive spacing around the header and final answer.

This needs a turn-level layout container that owns the same spacing contract, or explicit equivalent spacing for the reveal contents and header.

Relevant code:
https://github.com/tutti-os/tutti/blob/efe36b84e757777950d3e0e1aa90272b085d0e37/packages/agent/gui/shared/agentConversation/components/AgentTurnWorkSection.tsx#L84-L97

3. [P2] Do not suppress processing unless a renderable row can show canonical timing

agentProcessingProjection.ts removes the generic processing row whenever a canonical live turn has a finite startedAtUnixMs. However, the duration UI only exists when that turn has a rendered transcript group.

I reproduced a provider-initiated running turn with no user message and only a runtime diagnostic message that is later removed by dropCodexRuntimeDiagnosticNotices. The current projection returns an empty rows array: the generic processing row was suppressed, and no turn group remains to render the duration.

The suppression decision should happen after normalized rows confirm that the active turn has a row capable of hosting the duration, or canonical timing should be represented by an explicit row.

Relevant code:
https://github.com/tutti-os/tutti/blob/efe36b84e757777950d3e0e1aa90272b085d0e37/packages/agent/gui/shared/agentConversation/projection/agentProcessingProjection.ts#L13-L22

Verification: the current PR head passes the Agent GUI suite locally (182 test files, 1687 tests). I also added a temporary regression test for issue 3; it failed with expected one processing row but received an empty array. The temporary test was removed after verification.

hugozhou-ai (approved):
LGTM

PR #1571, opened 2026-07-23
fix(agent): 完善 KimiCode 模型、会话、Skill 与交互处理

Felix-cy110:
## Summary

- preserve KimiCode model switching, image-input projection, provider sessions, and prepared runtime context across daemon restarts
- materialize Tutti canonical Skills for KimiCode and correlate split ACP `session/request_permission` / `tool_call` frames before publishing canonical `ask-user` interactions
- translate AgentGUI answer labels back to provider-owned permission option IDs and fence response completion ahead of provider follow-up messages
- add a generic Agent Extension profile contract that keeps signed core slash commands under **Commands**, host browser support under **Capabilities**, and unlisted runtime-advertised entries under **Skills**
- keep Tutti-injected routing Skills out of the picker while preserving Kimi-native Skill names, descriptions, and exact slash triggers
- recover the historical Kimi Session snapshot shape whose provider is `acp:kimi-code` but whose provider-native fingerprint was computed by the older empty-provider algorithm
- document the extension composer projection and the historical Session diagnostic/recovery path

## Root cause

KimiCode exposes both core session commands and native Skills through ACP `available_commands_update`. Without an extension-owned projection contract, Tutti rendered every entry in the command section and could not produce the separate Command / Capability / Skill layout.

A separate transition bug affected historical Sessions. Some snapshots already retained the open extension provider (`acp:kimi-code`) while their provider-native fingerprint still used the previous empty-provider payload. The existing compatibility path handled snapshots where both the provider and fingerprint used the old empty value, but missed this mixed transitional shape, so resume failed with `session runtime snapshot is unavailable: provider-native fingerprint does not match`.

The compatibility read remains fail-closed: it applies only to an unregistered open provider when the canonical Session provider and snapshot provider match exactly and the fingerprint equals the historical empty-provider payload. The exact enabled Agent Target remains launch authority; the database is not rewritten.

## Ownership

This changes ACP transport/projection and `tuttid` adapter-side provider preparation. It does not define new Session, Turn, Goal, or runtime-operation lifecycle semantics, so no Agent Host API or conformance change is required.

## User impact

KimiCode's `/` palette now presents distinct **Command**, **Capability**, and **Skill** sections, including the host-managed Browser capability. Historical KimiCode conversations written during the provider-normalization transition can resume and accept new messages again.

## Validation

- Kimi extension package validation: 24/24 tests passed in the companion repository
- `go test ./service/agent -run 'SessionRuntimeSnapshot|PrepareRuntimeForResume'` — passed
- `go test ./service/agent ./service/agentextension` — passed
- staged formatting and repository boundary hooks — passed
- `pnpm check:changed` — 63/64 lanes passed; the remaining `packages/agent/daemon` Go lint lane reports pre-existing unused/deprecation findings from this branch's earlier ACP work and does not reference files changed by the final composer/history commit
- `git diff --check` — passed

## Documentation

Updated:

- `docs/architecture/agent-extensions.md`
- `docs/conventions/troubleshooting/agent-session-lifecycle.md`
- `docs/specs/2026-07-14-agent-extension-package-design.md`

## Checklist

- [x] This PR does not change Host-owned lifecycle semantics
- [x] The extension behavior is declarative and contains no provider-name branch in AgentGUI
- [x] Durable architecture and troubleshooting documentation is updated
- [x] README/CONTRIBUTING language variants are not affected
- [x] All new commits include DCO sign-off


services/tuttid/service/agent/composer_runtime_context.go
```diff
@@ -147,12 +162,81 @@ func (s *Service) composerRuntimeContextFromSession(
...
+		}
 		selectedUpdatedAt = updatedAt
 		selectedID = session.ID
 	}
 	return selected
 }
 
+// extensionSessionCapabilitiesFallback returns the non-empty capability list
+// advertised by the most recently created live or persisted session of the
+// same provider and agent target, ignoring installation, project scope, and
```
jomeswang: [Blocker] This fallback explicitly ignores `extensionInstallationId`, project scope, and settings signature, then reuses handshake capabilities from another session. That violates exact runtime identity: a reinstall/version change or a differently configured instance can advertise a different protocol surface, so the composer may expose image/plan/interrupt actions that the selected installation does not implement. Intersecting with manifest capabilities is insufficient because those are only an upper bound, not proof of the current runtime handshake. Please key persisted evidence by the exact installation/binding identity (and relevant config signature), and fail closed when no matching evidence exists.

packages/agent/runtimeprep/skill_templates/browser-use.kimi.md
```diff
@@ -0,0 +1,29 @@
+---
```
jomeswang: [High] The seven new `*.kimi.md` templates are currently unreachable. The embed glob includes them, but `provider_skill.go` still hardcodes only `skill_templates/browser-use.md`, `tutti-cli.md`, etc.; at this PR head there is no selector or reference to any `.kimi.md` filename. As a result these ~423 added lines cannot change Kimi runtime behavior and will drift from the canonical templates. Please either wire a provider-specific selection/materialization path with a test that asserts the emitted Kimi SKILL.md content, or remove these files from this PR.

packages/agent/daemon/runtime/acp_pending.go
```diff
@@ -64,6 +64,81 @@ func acpInteractiveResponseResult(action string, optionID string, payload map[st
...
+	}
+	switch normalizePermissionOptionToken(action) {
+	case "cancel", "cancelled", "dismiss", "dismissed", "skip":
+		return pending.resolvePermissionOptionID("deny")
+	default:
+		return "", false
+	}
+}
+
+func acpAskUserFirstAnswer(payload map[string]any) string {
```
jomeswang: [High] This silently collapses the canonical AskUser payload to one answer. With multiple questions, `answersByQuestionId` is ignored and the code falls back to `answers[0]`; with a multi-select answer, `acpAskUserAnswerText` returns empty unless the selection length is exactly one. The local interaction is then recorded as answered while the provider receives only one opaque option, so it can continue with incomplete/different input. If the Kimi bridge only supports a single single-select question, please validate and reject unsupported shapes instead of dropping data; otherwise preserve/map all answers and add multi-question + multi-select tests.

jomeswang (changes requested):
Requesting changes for 1 blocker and 2 high-priority issues: (1) capability fallback crosses installation/config identity and can expose stale unsupported runtime capabilities; (2) the added Kimi skill templates have no selection/materialization path; (3) AskUser adaptation silently drops multi-question/multi-select answers. Details and suggested fixes are in the inline comments.

packages/agent/runtimeprep/skill_templates/browser-use.kimi.md
jomeswang: After checking the actual runtime pinned by the extension (`@moonshot-ai/kimi-code@0.28.0`), I think the correct resolution is to **remove these `*.kimi.md` copies and reuse the canonical templates**, rather than add a provider-specific selector.

Kimi accepts the canonical directory-form `SKILL.md` frontmatter already present in the shared templates: `name` and `description` are sufficient; `type` defaults to `prompt`, `disableModelInvocation` defaults to allowing invocation, and `whenToUse` is optional. `installProviderNativeSkills` emits `<skill>/SKILL.md`, so the source template suffix has no runtime meaning.

The copies have also already drifted beyond metadata: some remove the canonical capability guards, several use a different placeholder dialect (`{{CLI_COMMAND}}` vs `{{.CLICommand}}`), and large templates such as `tutti-cli`/`issue-manager` have diverged substantially. Maintaining full provider copies would make future safety/command updates easy to miss.

The intended Kimi path should be:

`canonical *.md` → generic `ExtensionSkillRoots` materialization → workspace `.agents/skills/<name>/SKILL.md` → Kimi native project skill discovery.

Kimi Code 0.28.0 natively scans `<project-root>/.agents/skills`, including in ACP sessions via the session cwd. This avoids a Kimi-specific preparer. Its top-level `--skills-dir` flag should not be relied on here: the `kimi acp` handler currently does not propagate that option into the ACP harness, and the flag also replaces automatic user/project discovery rather than appending to it.

Please delete all seven `*.kimi.md` files, enable/decorate the Kimi extension with the verified `.agents/skills` composer root, and add an integration test that materializes a canonical template and confirms Kimi ACP `listSkills`/`available_commands_update` sees it.

services/tuttid/service/agent/composer_runtime_context.go
Felix-cy110: We acknowledge this issue. The root cause is that capabilities are split into SessionMetadata after persistence, while precise scope queries filter out capability-only sessions before the metadata is reassembled. Furthermore, the newly added fallback incorrectly bypasses verification of installation ID, project scope and settings identity.
For rectification, we will remove the cross-scope fallback logic. Instead, persisted capabilities will be reassembled only after an exact identity match is completed, with a fail-closed policy enforced when no matching evidence can be found. Regression tests covering reinstallation scenarios as well as mismatched project and settings configurations will also be added.

packages/agent/daemon/runtime/acp_pending.go
Felix-cy110: 确认该问题成立。Kimi Code 0.28.0 的 ACP AskUser 桥接目前只能返回单个 optionId，而 Tutti canonical payload 支持多题、多选。当前实现错误地回退到展示字段 answers[0]，会造成 provider 输入与本地记录不一致。我会改为仅使用 answersByQuestionId，显式校验单题单选能力，对多题、多选和自由文本 fail closed，并补充相应的生命周期与回归测试。

jomeswang (approved):

PR #1681, opened 2026-07-26
feat(agent): add durable Codex edit and retry

mt-hub8:
## Summary

- add a Host-owned durable Saga for editing and retrying only the latest completed Agent turn
- integrate Codex `thread/rollback`, authoritative `thread/read`, and replacement `turn/start` with stable idempotency identities and explicit recovery for unknown outcomes; treat a stable user-submit replay as idempotent when only transport sequence, source, or timestamps differ, while preserving content and semantics conflict checks
- keep retracted turns for audit while Session detail and authoritative history project only effective turns; preserve attachments and non-text input, and leave previously applied filesystem changes untouched
- route command state and authoritative message reconciliation through agent activity-core, removing a terminal optimistic row only when both its stable Turn and client-submit identity are absent from effective history, with a minimal Agent GUI edit-and-retry interaction
- document the lifecycle boundary, frontend data flow, effective-history projection, provenance replay semantics, and terminal-turn message reconciliation troubleshooting path

## Verification

- `pnpm check:changed` — all 44 planned lanes passed for `d5ff01b8d`, including lint, Go tests and lint, TypeScript typecheck, boundary checks, generated contracts, and changed TypeScript suites
- focused validation passed for SQLite submit-provenance replay and conflict rejection, tuttid effective-Turn detail projection, agent activity-core optimistic-history replacement (294 tests), and both Desktop edit-retry reconciliation scenarios

## Fallback Three Questions

- Source: Codex App Server exposes rollback and replacement start as separate RPCs, and a process or transport failure can leave either Provider acceptance or the local projection temporarily unknown.
- Why can it not be fixed at that source? Codex App Server does not currently provide one atomic edit-and-retry operation with a durable idempotency receipt, and SQLite plus the Provider cannot participate in one ACID transaction.
- Removal condition: replace the Host Saga only when the Provider supplies an atomic edit-and-retry API with durable idempotency and authoritative acceptance/history receipts; remove the terminal message reconcile fallback when terminal activity delivery guarantees a complete authoritative message projection.

## Checklist

- [x] If this PR changes agent lifecycle semantics (session/turn/goal/runtime-operation creation, sendability, terminal state, or recovery): the change lives in `packages/agent/host` and adds a `packages/agent/host/conformance` scenario, not in a `tuttid`/tsh adapter.
- [x] I updated all README/CONTRIBUTING language variants when changing their English source.


mt-hub8:
Confirmed against the current PR head (`5489063`). No further code change is needed here. A durably accepted `202` result no longer moves the reducer directly to `succeeded`: `settleCommand` stores the result and enters `reconciling`, and only an authoritative availability read with the matching history revision and recovery state settles the operation. The dispatch promise intentionally resolves at durable acceptance so the inline editor closes, while `resend_pending` / `recovery_required` are surfaced through the recovery presentation. This avoids a duplicate replacement submission while preserving an explicit durable recovery path.

mt-hub8:
Confirmed after auditing every production caller of `ReleaseOrFailRuntimeOperation`. Only edit-retry intentionally passes `NextAttemptAtMS == NowUnixMS`; the generic cancel / plan-decision / interactive paths still use `runtimeOperationNextAttemptAt`, which applies at least a one-second backoff, and failed operations do not retain a next-attempt timestamp. Claiming already uses `next_attempt_at_unix_ms <= now`, and the lease/CAS update still permits only one claimant. The immediate-release store test covers the equality contract. No current consumer depends on the former strict-after invariant, so no additional code change is needed.

PR #1695, opened 2026-07-27
fix(agent-gui): keep message locator in sync

rv4no:
## Summary
- pause message-locator and transcript virtualizer geometry work while AgentGUI is fully occluded, then restore bottom-locked virtual timelines on exposure
- make real wheel and keyboard scroll intent update locator selection immediately while still filtering layout-driven reverse movement
- restore the virtualized end when fast recovery replaces an empty timeline, with regressions for recovery, visibility, wheel reversal, and explicit scroll-to-end paths

## Tests
- `pnpm check:changed -- --push-ready`

## Notes
- user scroll-away bottom-lock behavior remains authoritative
- real Electron pixel verification was not run because no in-app browser runtime was available


rv4no:
最新 main 已合入统一的 transcript end-following 状态机（#1694）。这个 PR 基于旧 bottom-lock 结构继续叠加 virtualizer ownership 和 DOM event，现已冲突，不再适合作为修复基础。后续会基于最新 main 用单一 followEndMode/virtualizer owner 重新实现。

PR #1760, opened 2026-07-28
fix(agent): bound streamed tool output memory

jomeswang:
## Summary

- accumulate provider tool output deltas with a per-call `strings.Builder` instead of `current + delta`, eliminating full-prefix copying for every output chunk
- retain cumulative durable snapshots and byte-offset live append operations without changing the live EventHub protocol
- add explicit coverage that the `tool_call` coalescer now on `main` flushes a running snapshot before its terminal update, and document the complete memory/backlog failure pattern

A large streamed command currently rebuilds the full cumulative `output.text` for every provider delta. Even with the queue-level `tool_call` coalescing added by #1703, that source path remains quadratic in bytes copied and can create substantial allocation and compressor pressure. This PR makes source accumulation incremental; #1703 remains the owner of queue coalescing and terminal scheduling.

## Verification

- `pnpm check:changed` after rebasing onto current `origin/main` (7 lanes passed)
- pre-push `pnpm check:changed -- --push-ready` after rebase (8 lanes passed)
- focused iteration: `go test ./runtime -run TestStreamingReportCoalescer\|TestExplicitToolOutputDelta -count=1`

## Checklist

- [x] This does not change session/turn/goal/runtime-operation lifecycle semantics; it changes runtime output accumulation only.
- [x] I updated troubleshooting documentation for the affected data flow.
- [x] README/CONTRIBUTING translations are not affected.
- [x] The commit is signed off with DCO.

### DIFF
--- a/docs/conventions/troubleshooting/README.md
+++ b/docs/conventions/troubleshooting/README.md
@@ -27,11 +27,12 @@ Use the focused runtime index or open one area directly:
   startup that fails when concurrent provider input and output are treated as a
   strict scheduling order, false final-state mismatches caused by replay-generated
   child identities, and canonical completion delayed behind a streaming activity-report
-  backlog. It also covers provider-completed submissions reported as delivery
-  unknown after canonical message provenance conflicts, completed Claude Code
-  Turns that lack a Fork entry because provider identity was not observed from
-  the durable transcript, and Claude Fork operations that fail because an empty
-  SDK query never creates a durable provider child.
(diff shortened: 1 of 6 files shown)


jomeswang:
Closing because #1703 already fixes the incident’s retained-snapshot queue backlog. The remaining strings.Builder change is only a speculative allocation optimization without post-#1703 performance evidence; we can revisit it with a focused benchmark if needed.

PR #1770, opened 2026-07-29
fix(agent): settle Claude background task continuations

rv4no:
## Summary
- upgrade the Claude Agent SDK to 0.3.220 and treat `background_tasks_changed` as a replace-set level signal
- identify background follow-up results through `origin.kind = task-notification` and keep the root continuation open until authoritative session idle
- resolve every Claude tool event owner before closed-Turn admission and projection, so a settled child cannot suppress its parent Agent call completion
- fence background cleanup on cancel and failed root results, with timer and continuation state owned by `BackgroundTaskLifecycle`
- keep terminal child state monotonic and project terminal child summaries without stale `Starting…`

## Tests
- `pnpm check:changed -- --push-ready` (17 lanes passed)
- `node --test --experimental-strip-types packages/agent/claude-sdk-sidecar/src/sessionRuntime.delegated.test.ts` (23 tests passed)
- dev runtime verification: 6 delegated tasks completed, 0 failed/stopped, 10/10 root tool calls completed, and no `turn_completed_without_call_result`

## Notes
- delegation calls belong to the Session that launched the child; ordinary tools executed by a child belong to that child
- failed or canceled roots cannot be reopened by a late empty-level or task-notification signal
- no post-result settlement timeout is used because it can interrupt valid queued follow-ups and fabricate provider errors
- the existing 30-second timeout applies only when a reserved continuation never begins producing root output
- existing durable sessions are not backfilled


rv4no:
Addressed the active background-task review findings in `5903a79b9`:

- cancellation now clears the owned quiescence timer and pending continuation state before SDK shutdown
- failed or canceled root results fence late empty-level and task-notification continuation reservations
- added regressions for cancel-during-grace and both late-signal failure orders
- extracted `BackgroundTaskLifecycle`; `toolActivity.ts` is now 765 lines
- `claude_sdk_activity.go` is now 794 lines

Local pre-push validation passed all 17 changed-aware lanes.

PR #1775, opened 2026-07-29
feat(agent-gui): preserve session scroll position

Felix-cy110:
## What changed

- retain the scroll anchor and follow-end mode for each Agent Session visited by a mounted AgentGUI detail view
- restore detached DOM and TanStack Virtual transcripts to their previous offset when the user returns
- keep first-time and end-following sessions pinned to the latest turn
- document the UI-local ownership and lifetime of session scroll memory

## Why

Switching away from a Session discarded its viewport state. Returning to it therefore behaved like a fresh selection and jumped to the latest Turn instead of preserving the user's reading position.

## Impact

Users can move between Sessions without losing their place. The state remains local to the mounted AgentGUI view and does not change navigation, Engine, Session lifecycle, or persisted daemon state.

## Validation

- `pnpm check:changed -- --push-ready --base upstream/main` — 12 lanes passed
- focused AgentGUI scroll and virtual transcript tests — 58 tests passed
- pre-commit formatting, renderer/UI boundary, degradation, and runtime image budget checks passed

### DIFF
--- a/docs/architecture/agent-gui-node.md
+++ b/docs/architecture/agent-gui-node.md
@@ -901,10 +901,15 @@ High-frequency transcript updates must not pair DOM mutation with unconditional 
 Transcript end-following is one UI-local state machine shared by DOM, TanStack
 Virtual, and React Native adapters. It has only `following` and `detached`
 modes. User scroll-away intent detaches synchronously, before the first scroll
-frame. Conversation selection, prompt submission, an explicit scroll-to-end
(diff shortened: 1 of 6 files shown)


jomeswang (approved):
LGTM — scroll memory 生命周期绑定 useRef、writeTimelineScrollAnchor 封装一致、测试覆盖 detached/virtual/skeleton 三场景。小建议：scroll memory Map 可考虑后续加 LRU 上限，非阻塞。

PR #1887, opened 2026-07-30
fix(agent): neutralize durable edit-retry to stop boot crash loop

devRickyyy:
## Symptom
Users on 0.2.7-rc.0 hit a permanent **crash-on-open**: `tuttid` exits(1) on every launch with
`build tuttid server: recover agent host: process runtime operation <id>: agent session not found`,
and the desktop reports `failed to start managed tuttid: tuttid exited before it published its listener info.`

## Root cause
A durable `edit_retry` runtime operation that rolled back the last turn but never re-sent the
replacement (`ErrEditRetryResendPending`) is harmless while the daemon runs — the live worker just
logs and retries at 1 Hz. But on the next daemon restart the **cold-recovery pass**
(`RecoverRuntimeOperations`) treats the same error as **fatal**, aborting `build tuttid server`
before the listener is published. The operation is persisted in `workspace_agent_runtime_operations`,
so every relaunch re-reads it and re-crashes — a permanent loop for anyone who had a pending
edit-retry when they updated to a build containing #1681.

The upstream saga defect (why the op got stuck): the destructive `thread/rollback` runs before the
replacement is confirmed, the `ReplacementDispatched` checkpoint is a one-way fence that disables
auto-redispatch, and the replacement send is skipped when the submit-claim was already accepted — so
the edited turn is silently never resent, and the op hot-spins forever.

## Why not `git revert` #1681
It's a 183-file squash commit with ~37 conflicts against later work, and it ships a **destructive
one-way schema migration (v5)** that already ran on users' databases. Reverting the migration would
brick already-migrated installs a different way.

## This change — forward neutralization (no schema change)
Gate the feature behind `Config.EditRetryDisabled`, enabled in production via `composeApplicationHost`:
- `GetEditRetryAvailability` reports unsupported → the GUI hides the affordance
- `EditRetry` / `RecoverEditRetry` refuse before creating any operation
- the runtime-operation worker **quarantines** any leftover `edit_retry` op (marks it failed, returns
  nil) so it can neither crash cold recovery nor hot-spin the live worker

Existing poison pills **self-heal on the next boot** (they get quarantined during recovery). The zero
value keeps the feature enabled so its unit/conformance tests still exercise it; re-enable once the
saga's resend/recovery gap is fixed.

## Tests
- New `edit_retry_disabled_test.go`: builds a genuinely stuck op via the enabled saga, asserts
  recovering it with the feature **enabled is boot-fatal**, and asserts the **disabled path
  quarantines** it and returns nil (provider untouched).
- Full `packages/agent/host` module + tuttid edit-retry/conformance/api suites green; both modules vet clean.

## Operator rescue
`tools/scripts/rescue-edit-retry-poison-pill.sh` quarantines the stuck rows in `~/.tutti/tuttid.db`
for an already-bricked install that can't update yet (backs up first; refuses to run while tuttid is alive).

## Follow-up
A separate PR targets `release/0729` (its production host is composed in `host_adapters.go`, so the
wiring line differs).

---

<a href="https://app.devin.ai/review/tutti-os/tutti/pull/1887" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

---

> [!NOTE]
> **Medium Risk**
> Changes daemon boot recovery and disables a user-facing mutation path in production, but scope is bounded to edit-retry and avoids schema changes; incorrect quarantine logic could leave claimable poison pills or hide real recovery errors.
> 
> **Overview**
> Stops **tuttid** crash-on-launch when a stuck durable `edit_retry` runtime operation survives in SQLite and makes cold recovery fatal.
> 
> Adds **`Config.EditRetryDisabled`** on the agent host (production sets it in `composeApplicationHost`). When enabled: **`GetEditRetryAvailability`** reports provider unsupported so the UI hides edit-and-retry; **`EditRetry`** and **`RecoverEditRetry`** return **`ErrRuntimeHistoryUnsupported`** before creating or driving saga work; the runtime worker **`quarantineDisabledEditRetryOperation`** marks leftover `edit_retry` ops **failed** instead of running the saga, so boot recovery completes without touching the provider.
> 
> Adds **`edit_retry_disabled_test.go`** (refuse new ops; quarantine stuck op on recovery vs boot-fatal when enabled) and **`tools/scripts/rescue-edit-retry-poison-pill.sh`** to quarantine stuck rows in an already-bricked local DB after backup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f3e6ac9e983d52805645d95ce603a54a16776d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/conventions/troubleshooting/agent-session-lifecycle.md
+++ b/docs/conventions/troubleshooting/agent-session-lifecycle.md
@@ -290,8 +290,10 @@ incomplete`, while a newly created session can still launch.
   marks it failed AND clears the session's effective-history fence back to `ready`
   so the conversation can still send — and returns non-fatally, so existing poison
   pills self-heal on the next boot. New edit-retries are refused at the entry
-  points. The durable rule: never let one runtime operation's error abort daemon
-  startup.
+  points. Sessions fenced before the neutralization (owning operation already
+  failed, so recovery cannot see it) self-heal at the send gate: the first send
+  clears an abandoned fence instead of rejecting. The durable rule: never let one
+  runtime operation's error abort daemon startup.
 - **Rescue:** For an install that cannot update yet, quit Tutti and run
   `tools/scripts/rescue-edit-retry-poison-pill.sh`, which quarantines the stuck
   rows and clears the session fence in `~/.tutti/tuttid.db` after backing it up.
--- a/docs/conventions/troubleshooting/agent-session-lifecycle.md
+++ b/docs/conventions/troubleshooting/agent-session-lifecycle.md
@@ -268,6 +268,38 @@ incomplete`, while a newly created session can still launch.
   failure, second-scan non-duplication, and outcome-write failure stopping
   startup.
 
+### A stuck edit-and-retry operation crashes the daemon on every launch
+
+- **Symptom:** After updating, `tuttid` exits immediately on every launch and the
+  desktop reports `tuttid exited before it published its listener info`; the daemon
+  log shows `recover agent host: process runtime operation <id>: agent session not
+  found` (or `... agent history was rolled back but the edited turn still needs to
+  be resent`). The app never opens.
+- **Quick checks:** Look for a durable `edit_retry` runtime operation stuck at the
+  `resend_pending` checkpoint — the last Turn was rolled back on the provider but
+  the replacement was never re-sent. While the daemon runs, the live worker only
+  logs this at 1 Hz; the fatality appears only on the next cold restart. The
+  affected session's `workspace_agent_session_history.recovery_state` is fenced
+  (non-`ready`).
+- **Root cause:** The cold-recovery pass (`RecoverRuntimeOperations`) treats a
+  per-operation error as fatal to `build tuttid server`, so a persisted edit-retry
+  operation that can never make progress becomes a boot poison pill. The live
+  worker tolerates the same error, which is why the app worked until the restart.
+- **Fix:** Durable edit-and-retry is disabled (`Config.EditRetryDisabled`, set in
+  production wiring). Recovery quarantines any leftover `edit_retry` operation —
+  marks it failed AND clears the session's effective-history fence back to `ready`
+  so the conversation can still send — and returns non-fatally, so existing poison
+  pills self-heal on the next boot. New edit-retries are refused at the entry
+  points. The durable rule: never let one runtime operation's error abort daemon
+  startup.
+- **Rescue:** For an install that cannot update yet, quit Tutti and run
+  `tools/scripts/rescue-edit-retry-poison-pill.sh`, which quarantines the stuck
+  rows and clears the session fence in `~/.tutti/tuttid.db` after backing it up.
+- **Validation:** `packages/agent/host/edit_retry_disabled_test.go` builds a
+  genuinely stuck operation, asserts enabled recovery is boot-fatal, and asserts
+  the disabled path quarantines it, returns nil, and leaves the session at
+  `recovery_state = ready`.
+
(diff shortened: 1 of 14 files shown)


devRickyyy:
Addressed the review findings in `4909fa5`:

**🔴 Quarantine left sessions send-blocked (cursor + devin).** Correct — marking only the runtime-operation row failed left the session's `workspace_agent_session_history.recovery_state` fenced at `resend_pending`, and with the feature disabled no path returned it to `ready`, so `requireSendAllowedByEffectiveHistory` rejected every subsequent send. Fixed by adding `storesqlite.QuarantineEditRetryOperation`, which **atomically** fails the operation *and* clears the history fence back to `ready` (dropping `operation_id`), mirroring `AbortEditRetryRollback` but tolerating any fenced state and abandoning the retracted turn so the conversation stays usable. Wired through the `EffectiveHistoryStore` port, observed store, and workspace adapter. The regression test now asserts the session is at `resend_pending` before quarantine and `ready` after. The rescue script was updated with the same fence-clearing `UPDATE`.

**🟡 Docs not updated (devin).** Added a troubleshooting entry ("A stuck edit-and-retry operation crashes the daemon on every launch") to `docs/conventions/troubleshooting/agent-session-lifecycle.md`, and a note in `packages/agent/host/README.md` that edit-and-retry is currently disabled via `Config.EditRetryDisabled`.

Backported to `release/0729` in #1888.

devRickyyy:
Follow-up in `b855e2b` (self-identified gap, not from review): the quarantine path only covers **claimable** operations, which recovery drains at boot. A session fenced *before* the neutralization — its operation already `failed` (e.g. via `FailEditRetryRecovery`, fence at `recovery_required`) — is invisible to recovery, so its fence survived every boot and the session kept rejecting all sends.

Fix: `storesqlite.ClearAbandonedEditRetryFence` clears a session's fence when the owning operation is missing or terminal (in-flight fences are left to the worker's quarantine), and `requireSendAllowedByEffectiveHistory` attempts this heal before rejecting when the feature is disabled — so the first send on an orphan-fenced session proceeds instead of failing forever. Implemented at the send gate rather than a boot sweep because the `EffectiveHistoryStore` port is workspace-scoped and workspace DBs may attach lazily.

New regression test reproduces the legacy terminal state, guards that the enabled gate rejects it (proving the test reaches the fence), and asserts disabled recovery leaves it untouched while the first disabled send heals it to `ready`. Cherry-picked to #1888 as `49a920e`.

devRickyyy:
Scope reduction in `e16f440` (revert of `b855e2b`), applying a stricter conservatism rule for this hotfix: **every line of the fix must execute only on installs actually affected by the incident — no daily code path may change.**

The send-gate lazy heal violated that: it modified `requireSendAllowedByEffectiveHistory` — the gate on every send of every session — turning a pure-read check into a conditional write on the hottest user path. And its beneficiaries (sessions fenced at `recovery_required` by an operation that failed *before* the neutralization) were a **pre-existing dead end**: that state offered no recovery action even when the feature was enabled, so it is not a regression of this PR. Those installs are covered by the rescue script, and the troubleshooting entry now says so explicitly.

Final shape of the PR:
- entry gates — only touch the disabled feature
- boot quarantine (atomic op-fail + fence-clear) — only runs when a poison-pill operation exists
- send path — **byte-identical to before the PR**

Mirrored on #1888 as `ff5ee27f`.

PR #1991, opened 2026-08-03
feat(agent): stabilize native Codex skill roots

jomeswang:
## Summary

- enable the existing content-addressed managed-Skill store for `local:codex`
- configure native Codex through `skills/extraRoots/set` after `initialized` and before `thread/start` or `thread/resume`
- snapshot provider-owned `.system` Skills into a stable content-addressed path while preserving the existing Tutti Agent digest/marker contract
- migrate legacy run-scoped managed Skills safely, including root-symlink containment
- raise the native Codex minimum version to `0.136.0`, the first release with `skills/extraRoots/set`

No external `tutti-agent` repository change is required; native Codex and Tutti Agent continue to share the same app-server adapter behind provider registry strategies.

## Validation

- `pnpm check:changed -- --push-ready`: 13 lanes passed
- focused Go suites: 1,459 tests passed across runtimeprep, runtime, and providerregistry
- live native Codex 0.144.5 protocol test: stable managed, user, and system Skill paths across distinct `CODEX_HOME` values
- independent review completed; two migration/symlink containment findings were fixed and re-reviewed with no blocking findings

### Controlled cross-session cache result

Using two fresh `local:codex` sessions with identical `hi`, model, cwd, permissions, and startup context:

| Build | Session 1 | Session 2 |
| --- | ---: | ---: |
| `origin/main` | 9,984 / 16,285 cached (61.31%) | 9,984 / 16,285 cached (61.31%) |
| this branch | 9,984 / 16,285 cached (61.31%) | 9,984 / 16,285 cached (61.31%) |

The patch makes the normalized AGENTS, Skill sources, system Skill target, and non-assistant message content identical across sessions, but measured cached-token improvement is **0** in this controlled run. Native Codex currently derives `prompt_cache_key` from its fresh thread id and does not expose an app-server override, matching upstream report https://github.com/openai/codex/issues/21796. Therefore this PR establishes stable Skill inputs and the RPC boundary, but does not claim cross-thread billing/cache improvement until Codex exposes stable cache affinity.

## Compatibility and rollback

- existing Tutti Agent digest prefix, marker, and cache schema remain unchanged
- legacy native Codex managed Skills are removed only when their ownership marker is a regular file; unowned directories are preserved
- a linked `skills` root is unlinked without following its target, then recreated inside the run root
- rollback is the single commit in this PR; stable bundle directories are rebuildable cache data and may remain harmlessly on disk


jomeswang:
Closing because the controlled cross-session A/B showed no cache improvement: origin/main and this change both reported 9,984 cached input tokens out of 16,285 (61.31%). The stable Skill roots/prefix behavior was verified, but it did not produce measurable cache benefit for native Codex. The remote branch is intentionally retained so this can be revisited if upstream exposes a stable cross-thread prompt_cache_key or new provider-side evidence becomes available.

PR #2009, opened 2026-08-04
feat(agent): add turn performance telemetry and live processing status

Liyu199808:
## Summary

Adds provider-neutral Agent Turn observability and clearer waiting feedback.

- Reports one `agent.turn_performance` terminal summary for each canonical Turn.
- Captures TTFT, first progress, total duration, provider, model, Session state, outcome, idle gaps, queue state, and tool usage when reliably available.
- Measures TTFT and total duration from the client submit boundary when the
  privacy-reviewed submission provenance is available.
- Preserves only `clientSubmittedAtUnixMs`, `sessionState`, and `queued`
  through the durable submit-provenance path instead of forwarding arbitrary
  submission metadata.
- Uses a strict content-free analytics whitelist. Prompt text, responses,
  reasoning, file contents, paths, commands, tool arguments, URLs,
  authentication values, and local business identifiers are not included.
- Keeps analytics reporting asynchronous and best-effort so reporting failures
  cannot block Agent execution.
- Adds one stable localized processing row for preparing, submitting, waiting,
  reasoning, generation, tool execution, tool result waiting, and reconnecting.
- Uses generic waiting copy when the real Agent phase cannot be confirmed.
- Omits Token, retry, and reconnect values when no reliable Turn-scoped source
  exists.
- Updates the AgentGUI and analytics architecture documentation.

## Verification

- `pnpm --filter @tutti-os/agent-gui test`
  - 300 test files passed
  - 2579 tests passed
- AgentGUI typecheck passed.
- `pnpm check:i18n` passed.
- `pnpm check:agent-host-boundary` passed.
- `pnpm check:agent-activity-runtime-boundaries` passed.
- `pnpm check:ui-boundaries` passed.
- `pnpm check:agent-gui-degradation` passed.
- Full relevant Go packages passed:
  - `packages/agent/host`
  - `packages/agent/daemon/runtime`
  - `packages/agent/daemon/hostadapter`
  - `services/tuttid/service/agent`
  - `services/tuttid`
- Added coverage for new/existing Session provenance, known queue states,
  runtime-first/provenance-first persistence order, client-submit timing, and
  terminal-event privacy filtering.
- `git diff --check` passed.

The Runtime full test suite requires localhost `httptest` binding. It failed in
the managed sandbox only because port binding was denied, then passed when
rerun with local test-port access.

Manual verification before marking the PR Ready:

- New Session: `timing_start_source=client_submit`, `session_state=new`.
- Existing Session: `timing_start_source=client_submit`,
  `session_state=existing`.
- Immediate request: `was_queued=false`.
- Exactly one terminal `agent.turn_performance` event is emitted.
- The event contains no Prompt, response, reasoning, path, command, tool
  argument, complete URL, authentication value, workspace ID, Agent Session
  ID, Turn ID, or submit ID.
- The processing row stays on one line and disappears at the terminal state.

## Fallback Three Questions

- Source: some providers do not expose a uniform phase, heartbeat, Token usage,
  retry count, or reconnect count while a Turn is active.
- Why can it not be fixed at that source: Tutti supports multiple providers
  with different runtime protocols, and there is currently no provider-neutral
  structured contract for all of these facts.
- Removal condition: remove the idle presentation fallback when all supported
  providers expose a normalized progress/phase heartbeat. Report Token, retry,
  and reconnect values only after a reliable Turn-scoped provider-neutral
  source is available.

## AgentGUI Review

| Lane | Trigger | Result | Evidence | Affected consumers |
| --- | --- | --- | --- | --- |
| Architecture | Turn lifecycle, projection, controller, and runtime state are observed | Pass | Canonical committed Turn owns the terminal event; Engine/runtime facts feed a pure UI projection; architecture and boundary checks passed | Desktop workspace AgentGUI, standalone Agent window, tuttid |
| Performance | Adds a one-second clock, idle fallback, terminal analytics read, and bounded asynchronous reporting | Pass | One shared clock drives the stable row; reporting is detached with a bounded timeout; degradation checks and relevant tests passed | AgentGUI transcript consumers, tuttid |
| Consumers | Changes shared processing-row projection and Host/runtime submit provenance | Pass | Optional provider-neutral facts fail closed; AgentGUI, i18n, boundary, Host, Runtime, adapter, and service tests passed | Desktop AgentGUI, standalone Agent window, shared AgentGUI hosts |

## Remaining Risks

- Token, retry, and reconnect metrics remain unavailable until reliable
  Turn-scoped structured sources exist; unavailable values remain null rather
  than reporting fabricated zero values.
- DataFinder must register `agent.turn_performance` and its event attributes
  before production rollout.
- Production DataFinder visibility requires a released production build;
  development builds report only to the local analytics debug panel.
  
  本次运行看到这个埋点数据：
img_v3_02148_10f4855d-cdfe-42c2-a438-c840161feb0g


devRickyyy:
**P0 blocker: `reported` dedupe state is unbounded and retained for the daemon lifetime**

`agentTurnPerformanceState.reported` is a `map[string]struct{}` that receives one key for every claimed terminal Turn. However, the 6-hour TTL pruning and 4,096-entry cap are only applied to `provenance`; `reported` is never pruned or evicted.

There is an additional amplification: `claimTurnPerformanceReport()` inserts into `reported` before checking whether `analyticsReporter` is available and before the repository reads / `Track()` call. Therefore, a disabled reporter or any later reporting failure still retains the key permanently.

Impact: memory usage in a long-lived `tuttid` process grows with the total number of terminal Turns, contradicting the stated bounded in-memory design.

Please make the reported/claimed state bounded as well (for example, store timestamps and apply the same TTL + capacity policy, or use a bounded TTL cache), avoid claiming when reporting is disabled, and add coverage for TTL/capacity plus failed-report dedupe semantics.

devRickyyy:
### Concern: the telemetry path should not invoke the production model catalog

The current terminal-turn reporting path is:

`reportAgentTurnPerformance` → `resolveAgentTurnAnalyticsModel` → `catalog.ListModels`

This call is only being used to normalize `Session.Model` into a telemetry value (known model / `custom` / `unknown`), but it couples analytics reporting to the production catalog implementation.

For every eligible terminal Turn with a non-empty model, the reporter calls `ListModels`. A cache may avoid doing the underlying work every time for some providers, but:

- Codex and Tutti Agent use a short catalog cache (30s; 5s on errors).
- OpenCode currently has no catalog TTL, so a report can execute `opencode models --verbose` for each Turn.
- There is no singleflight around a cold cache, so concurrent Turn completions can start duplicate catalog loads.
- Depending on the provider, loading the catalog may start a CLI/app-server process, read runtime configuration, or prepare/refresh the model-catalog auth view.

That gives a best-effort telemetry path unnecessary CPU/I/O and authentication side effects. It can also make telemetry loss provider- and workload-dependent, because the entire report has a 3-second deadline.

**Suggested direction:** capture an already-known, low-cardinality provider/model classification in the Turn submission provenance (for example built-in model ID, `custom`, or `unknown`) and report that value after settlement. Terminal telemetry should not query the live production catalog. If catalog-derived metadata is needed, expose it from the existing model-selection/provider-preparation flow instead of loading the catalog again from the reporting path.

---

**Update — fixed in [`cc3858e9a`](https://github.com/tutti-os/tutti/commit/cc3858e9a69425aa183fdc9527bbd8d987e1ad64):**

- `agentRuntimeAdapter.Exec` captures the existing runtime Session provider/model before provider dispatch and stores it in the same bounded, process-local Turn provenance.
- Terminal reporting normalizes that snapshot directly. The model-catalog injection and `catalog.ListModels` call have been removed, so telemetry cannot start model discovery/provider processes.
- After daemon restart or provenance eviction, reporting falls back to the canonical Session provider/model without querying the catalog.
- Regression tests and the analytics architecture documentation now cover this boundary.

PR #2146, opened 2026-08-08
fix(desktop): capture composer follow-ups + configurable capture shortcut

devRickyyy:
## Summary

Follow-ups to the screenshot capture composer (#2109): three user-reported fixes plus a configurable capture shortcut.

- destroy the capture window after a confirmed submission so it can no longer stay on screen after send
- open the workspace Agent GUI on the newly created session when the workspace window is revealed after submission
- keep the capture footer controls on the documented single alignment baseline by hiding the connector control Quick Composer hosts cannot service
- make the global capture shortcut a durable preference, adjustable under Settings → General

### Submitted capture window could stay on screen

The submit handler tore the window down with `close()`, which a renderer
lifecycle handler can hold open — the exact stranding that already moved
cancellation to `destroy()` inside this feature. The submit success path now
mirrors `cancelCapture`: mark `closing`, then `destroy()`. Failed activation
still keeps the composer open with its draft intact, and the `closed`-event
placement persistence keeps firing.

### Revealed window did not navigate to the created session

Submission only focused the workspace window; nothing told the Agent GUI which
session to open, so the revealed window stayed on whatever it showed last. The
external `agentActivity.activateSession` contract gains an optional `reveal`
flag (normalizer-validated, absent means the existing activate-without-
navigation behavior). Capture submission requests it, and the workspace
external bridge then routes `requestWorkspaceAgentGuiLaunch` to the activated
session with the provider taken from the activation result — covering both the
workbench window and the standalone Agent window handlers. Navigation is
best-effort: a failed launch never fails the activation result, so the capture
window cannot be stranded open for a session that was actually created.

### Footer controls wrapped onto a second line

The connector capability control rendered inside the capture composer as a
dead trigger: Quick Composer hosts expose no capability-settings channel, so
it could never be serviced, and its width pushed the right-hand controls onto
a second row below the documented one-baseline toolbar. Quick Composer now
pins `connectors: { enabled: false }`; the primary-capability slot falls back
to the Tutti Mode chip, which renders nothing without a host activation
handler.

## Feature: configurable capture shortcut

Settings → General gains a shortcut recorder row for the screenshot capture
binding.

- The binding persists as a new `captureScreenshot` field on the
  workbench-shortcuts preference. It is optional on the wire (older clients
  stay valid) and required in the desktop shared type so every construction
  site is compiler-enforced. Unlike the other two bindings, **null means the
  built-in `CommandOrControl+Shift+S` default applies, not "unbound"** —
  clearing the row restores the default; capture can never be disabled by
  preference.
- The field flows through the daemon: OpenAPI schema + regenerated Go/TS
  clients, the biz model and normalizer, SQLite persistence, and the
  `preferences.desktop.updated` event payload including its strict topic
  schema (`desktop-preferences.schema.json`), which would otherwise drop the
  whole publish once the field appears.
- The capture service resolves its accelerator from the preference, subscribes
  to host preference changes, and re-registers idempotently. The recorder and
  the main-process resolver both require a Meta/Ctrl/Alt modifier so a global
  accelerator can never shadow plain or shift-only typing, and a binding
  Electron rejects (malformed or OS-conflicted) falls back to the last working
  accelerator with a logged warning.

`docs/architecture/desktop-screenshot-to-qute.md` is updated in place for all
of the above.

## Testing

- `packages/workspace/external-core`: `node --test` — 33 pass (adds `reveal`
  round-trip, omission, and rejection cases)
- `packages/agent/gui`: `vitest run AgentGUIQuickComposer.spec.tsx` — 17 pass
  (adds connectors-trigger-absent assertion)
- `apps/desktop`: capture + preferences/settings node tests — 69 pass,
  including new `resolveCaptureAccelerator` cases (valid bindings, arrow-key
  mapping, shift-only/malformed fallbacks)
- Go: `services/tuttid` biz/preferences, data/workspace (SQLite round-trip now
  covers `captureScreenshot`), service/eventstream — pass
- `generate:api --check` and `generate:event-protocol --check` clean; `tsgo`
  typecheck clean on all touched packages

Manual verification checklist (capture needs the global shortcut plus macOS
screen-recording consent, so this stays manual):

1. `Cmd+Shift+S` → select region → send → the floating composer disappears
   immediately
2. the workspace/Agent window that comes forward is showing the newly created
   session
3. the footer renders `+ @ agent target / project selector … permission,
   model, Create-Task switch, send` on a single row with no connector button
4. set a custom shortcut in Settings → General → the old combo stops working,
   the new one captures; restart the app → the custom shortcut still works
   (exercises PUT → SQLite → reload end to end)

### DIFF
--- a/apps/desktop/src/main/capture/captureShortcut.test.ts
+++ b/apps/desktop/src/main/capture/captureShortcut.test.ts
@@ -2,10 +2,32 @@ import assert from "node:assert/strict";
 import test from "node:test";
 import {
   desktopCaptureAccelerator,
-  desktopCaptureShortcutKeys
+  desktopCaptureShortcutKeys,
+  resolveCaptureAccelerator
 } from "./captureShortcut.ts";
 
 test("desktop capture uses a discoverable shortcut with at most three keys", () => {
   assert.equal(desktopCaptureAccelerator, "CommandOrControl+Shift+S");
   assert.ok(desktopCaptureShortcutKeys.length <= 3);
+});
+
+test("capture accelerator follows a valid persisted binding", () => {
+  assert.equal(resolveCaptureAccelerator("Meta+Shift+P"), "Meta+Shift+P");
+  assert.equal(resolveCaptureAccelerator("Ctrl+Alt+F5"), "Ctrl+Alt+F5");
+  assert.equal(resolveCaptureAccelerator("Alt+Space"), "Alt+Space");
+  assert.equal(resolveCaptureAccelerator("Meta+ArrowUp"), "Meta+Up");
+});
+
+test("capture accelerator falls back to the default for unusable bindings", () => {
+  assert.equal(resolveCaptureAccelerator(null), desktopCaptureAccelerator);
+  assert.equal(resolveCaptureAccelerator(""), desktopCaptureAccelerator);
+  assert.equal(resolveCaptureAccelerator("  "), desktopCaptureAccelerator);
+  assert.equal(resolveCaptureAccelerator("S"), desktopCaptureAccelerator);
+  // Shift-only bindings would shadow plain typing system-wide.
+  assert.equal(resolveCaptureAccelerator("Shift+S"), desktopCaptureAccelerator);
+  assert.equal(
+    resolveCaptureAccelerator("Meta+Shift"),
+    desktopCaptureAccelerator
+  );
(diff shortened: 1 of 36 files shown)


devRickyyy:
Superseded by #2191, which now targets `main` and contains the complete three-PR stack.

PR #2150, opened 2026-08-08
refactor(agent-gui): extract composer settings core and adopt in capture

devRickyyy:
## Summary

Root-cause fix for the capture panel's model-selection bugs, delivered as the first slice of the composer-settings policy extraction. Stacked on #2146 (base is its branch; GitHub retargets to main when it merges).

- new `@tutti-os/agent-gui/composer-settings-core`: headless, host-agnostic composer settings policy (ports & adapters + external-store contract)
- new `agentActivity.rememberComposerDefaults` external operation routing launcher picks into the canonical per-target composer-defaults ledger
- the capture window controller adopts the core; its hand-rolled fencing, failure semantics, and settings plumbing are deleted

## Why

The capture panel reused the Composer view but re-implemented the settings policy, and drifted:

1. an untouched panel submitted **no settings at all**, so the daemon seeded the model from the per-target defaults ledger (or catalog default) — a model the panel never displayed;
2. a failed or slow (~8s catalog probe) options refresh **blanked the model menu** and hard-disabled controls, silently swallowing picks and Enter;
3. picks lived only in a per-capture in-memory draft — never persisted, reset on every capture.

The agent GUI never had these bugs because its display and creation both read/write the same defaults ledger. This slice makes capture isomorphic to it.

## Design

`ComposerSettingsCore` (pure TS, no React/Electron; mirrors the `AgentSessionEngine` ports pattern and the repo's external-store conventions):

- **Fenced options lifecycle**: revision fence in a pure reducer; a slow stale response (the exact interleave from the incident logs) settles nothing. Regression-tested first.
- **Last-good retention**: a failed refresh keeps the previous menu and capability verdict (`degraded` flag instead of blanking); only the very first load for a target locks settings/submit.
- **Defaults seeding & write-back**: display seeds from the daemon's defaults-merged `effectiveSettings`; explicit picks write back to the same ledger through `rememberComposerDefaults` with trailing-edge coalescing (one in-flight write, latest patch wins, dropped on target change).
- **Display-what-you-send**: `resolveSubmitSettings()` returns exactly the displayed values (draft ⊕ effective); submissions always carry them, so the daemon never re-interprets empty fields against another surface's memory.
- Target change resets draft+options (per-target facts); cwd change keeps the draft.

Capture adoption details:

- capabilities now derive from the fenced options snapshot in the merged view snapshot — the previous refresh side-channel (which could apply a stale capability verdict) is gone;
- the main process logs submitted settings (`screenshot capture submitted`) — the panel previously had zero durable record of what it displayed;
- `resolveCaptureAgentCapabilities` moved to `src/shared/` so main and renderer share one derivation.

The bridge/guest transport legs mirror the existing `userProjects.*` proxy pattern (contract + normalizer + preload channel + main handler + bridge case).

## Testing

- `composer-settings-core` unit tests (7): stale-response fencing (the incident regression), last-good on failure, draft-over-effective resolution, target/cwd reset semantics, defaults-write coalescing and target-change drop, snapshot reference stability
- capture controller tests (rewritten, +2): submission carries displayed effective settings **without any pick** (the mismatch regression); failed refresh keeps menu+capabilities interactive
- external-core normalizer tests for the new operation
- full `agent-gui` vitest (2715), desktop capture node tests (26), typecheck clean on agent-gui / external-core / desktop; pre-commit ratchets green (the agent-gui degradation checker flagged and shaped the core's diagnostics port)

## Follow-ups (out of slice)

- **initial-load stall**: the core makes the first ~8s options load non-destructive but not invisible — a fresh capture window still locks settings/submit until the daemon catalog probe below is cached
- daemon: model-catalog probe single-flight + TTL cache (removes the recurring ~8s stalls); fail-closed create validation when the catalog is unavailable; log requested model alongside resolved in the create trace
- adopt the core in the main agent GUI composer and mobile (`composerDraftService`) — deletes the remaining parallel draft implementations
- capture window diagnostics channel (the core's `reportDiagnostic` port exists but no capture host feeds it yet)


devRickyyy:
Superseded by #2191, which now targets `main` and contains the complete three-PR stack.

PR #2386, opened 2026-08-14
feat(connectors): implement control plane v2

chovy-ai:
## Summary

Connector catalog refresh previously entered the lifecycle operation manager with an empty connector key, so it could conflict with unrelated installs. This breaking v2 change separates catalog synchronization from connector lifecycle work and makes operation recovery explicit.

- replace incremental refresh operations with a versioned Last-Known-Good catalog snapshot
- add long-poll delivery with immediate first pull and a 10-minute fallback
- use connector-scoped revisions and explicit conflicting lock claims
- persist accepted, running, outcome_unknown, recovering, completed, and failed operation states
- use a fresh v2 SQLite schema with no v1 compatibility path
- fail closed on revocation and schedule deterministic automatic uninstall
- expose only the local Connector Market v2 API

## Verification

- node tools/scripts/generate-openapi.mjs --check
- go test ./packages/connector/host ./packages/connector/daemon ./packages/connector/store-sqlite ./packages/connector/runtime ./services/tuttid/api
- pnpm --filter @tutti-os/connector-market typecheck
- pnpm --filter @tutti-os/connector-market test (57 passed)
- pnpm --filter @tutti-os/client-tuttid-ts typecheck
- pnpm --filter @tutti-os/desktop typecheck
- pnpm check:changed -- --push-ready (30 lanes passed)

## Fallback Three Questions

- Source: long-poll delivery can be interrupted while a desktop is asleep, offline, or reconnecting.
- Why source cannot fully fix it: the current delivery channel does not provide a durable per-device acknowledgement cursor.
- Removal condition: remove the 10-minute fallback after delivery provides durable per-device cursors with monitored completeness.

## Checklist

- [x] If this changes agent lifecycle, platform behavior, safety boundaries, or execution guarantees, I added or updated regression coverage.
- [x] I ran the focused tests/checks for the changed packages.
- [x] I updated architecture or protocol documentation where applicable.
- [x] I checked README variants; no translated README update is required.
- [x] I ran repository push-ready checks.
- [x] The commit includes a DCO Signed-off-by trailer.

### DIFF
--- a/apps/desktop/src/renderer/src/features/connector-market/services/internal/desktopConnectorMarketBackend.test.ts
+++ b/apps/desktop/src/renderer/src/features/connector-market/services/internal/desktopConnectorMarketBackend.test.ts
@@ -10,6 +10,7 @@ import { createDesktopConnectorMarketBackend } from "./desktopConnectorMarketBac
 test("desktop connector market backend delegates snapshot reads to the daemon client", async () => {
   let calls = 0;
   const snapshot: ConnectorMarketSnapshot = {
+    contractCohort: "connector-control-plane-v2",
     catalogState: "ready",
     connectors: [],
     operations: [],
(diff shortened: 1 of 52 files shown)


chovy-ai:
Related rollout PRs: tutti-lab/tsh#2727 and tutti-lab/tsh-server#821.

PR #2446, opened 2026-08-17
fix(agent-gui): restore manual unread for historical sessions

jomeswang:
## Summary

Restore Agent GUI's context-menu "Mark as unread" action for historical sessions.

## Root cause

The regression was introduced by `ead4da1143715fea5235af1d4208d8aff44d7a6` (`fix(agent-gui): preserve unread for unfocused sessions (#2372)`). That change removed the historical attention snapshot fallback, while `attention/unreadRequested` still looked for `sessionsById[id].latestTurn`. Production canonical sessions intentionally omit `latestTurn`; the turn is stored in canonical `turnsById`.

Runtime A/B evidence:
- `origin/main`: historical session snapshot + `attention/unreadRequested` produced no attention record.
- Parent `8d64f889...`: the same sequence created the historical attention record and manual unread state.

## Changes

- Resolve the latest canonical turn from `turnsById` for manual unread requests.
- Normalize session ids consistently before canonical lookup.
- Persist a manual unread record created before first read-state hydration.
- Extract and directly test canonical turn ordering.
- Add root-reducer and Agent GUI context-menu regression coverage.

## Verification

- activity-core: 489/489 tests passed.
- Agent GUI targeted regression: 20/20 tests passed.
- activity-core typecheck: passed.
- agent-gui typecheck: passed.
- oxfmt check and `git diff --check`: passed (only Windows LF/CRLF notices).
- Desktop bundle build from this worktree: passed.
- Independent subagent review: APPROVE; review follow-ups were applied and re-tested.

## E2E status

The replay reached the real Desktop/daemon/renderer startup path, but semantic assertions were blocked by stale cassette/provider protocol data: cassette expected ACP `turn/start` id `7`, while the runtime emitted id `8`, ending in `ReplayDeadlineExceeded`. This is an infrastructure/cassette compatibility issue, not a manual-unread assertion failure.

## Risk

The production change is isolated to activity-core attention state and reuses canonical turn entities already owned by the Engine. The remaining validation risk is that the existing replay cassette must be refreshed or made protocol-compatible before a full semantic E2E assertion can run.

### DIFF
--- a/packages/agent/activity-core/src/engine/agentActivitySnapshot.projector.test.ts
+++ b/packages/agent/activity-core/src/engine/agentActivitySnapshot.projector.test.ts
@@ -5,6 +5,7 @@ import {
   createInitialAgentSessionEngineState,
   rootEngineReducer
 } from "./rootReducer.ts";
+import { canonicalTurnKey } from "./sessionEntityKeys.ts";
 import type { AgentActivitySession } from "../types.ts";
 
 test("projects canonical engine state and preserves the snapshot reference", () => {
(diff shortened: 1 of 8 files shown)


jomeswang:
## dev-gui startup verification

Verified from commit `98a99c8f6f0d0bdf0d0483cc5a5ccd871e3526c9` in the isolated worktree on Windows.

- The repository has no native `make`; executed the exact `dev-gui` target body, `bash ./tools/scripts/dev-gui.sh`, under Git Bash with the repository-required Go 1.24.3.
- Built onboarding assets, the daemon, CLI, Electron main/preload, and warmed the renderer: 2325 modules.
- Electron started from this worktree and spawned `C:\Work\wt-agent-gui-manual-unread-20260818\apps\desktop\build\tuttid\tuttid.exe`.
- The daemon wrote its listener artifact at `C:\Users\15514\.tutti-dev\run\tuttid.listener.json` and listened on `127.0.0.1:51464`.
- Authenticated `GET /v1/health` returned HTTP 200 with `{"service":"tuttid","status":"ok"}`; `GET /v1/workspaces` also returned HTTP 200.
- Final desktop log recorded `managed tuttid spawned`; there was no `spawn UNKNOWN`, `tutti-desktop-startup-failed`, or bootstrap failure after the path-normalization fix.

The earlier `spawn UNKNOWN` was reproduced before this commit and was caused by passing a Git-Bash `/c/...` path to Node's Windows `spawn`; this commit normalizes the daemon path for the Node/Electron process. The dev-gui process tree was stopped after verification.

PR #2484, opened 2026-08-19
fix(desktop): hide release notes action on release builds

jomeswang:
## What changed

- Gate the desktop `更新内容` / `What's new` action with a build-time flag.
- Keep the action enabled for builds triggered from `main`.
- Disable it for `release/*` and tag builds on macOS, Windows, and Microsoft Store packages.
- Keep local development enabled by default.

## Why

Release-branch packages should not expose the release-notes entry, while `main` RC/validation builds should keep it available.

## Validation

- Targeted app-update presentation tests: 2 passed.
- Desktop typecheck: passed.
- Prettier and diff checks: passed.
- Desktop renderer build with the release flag disabled: passed; 8,086 modules transformed and renderer CSS contracts passed.
- Verified the compiled renderer contains `releaseNotesActionEnabled = false` under the release build flag.
- Independent review completed; Store workflow omission was fixed before updating this MR.

## Note

The local pre-commit/pre-push hooks also report a pre-existing stale renderer-token state in `origin/main`; no token files are included in this change.

### DIFF
--- a/.github/workflows/desktop-release.yml
+++ b/.github/workflows/desktop-release.yml
@@ -401,6 +401,7 @@ jobs:
           APPLE_API_KEY_ID: ${{ secrets.APPLE_API_KEY_ID }}
           APPLE_API_ISSUER: ${{ secrets.APPLE_API_ISSUER }}
           NODE_OPTIONS: --max-old-space-size=4096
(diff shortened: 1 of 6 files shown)


jomeswang:
本次需求应只落在 release/0818，原 MR 目标分支错误，已关闭；后续会从 origin/release/0818 重新创建正确基线的 MR。

PR #2505, opened 2026-08-19
fix(desktop): keep installed apps out of dock

jomeswang:
## Summary

- make installed workspace apps use the shared when-open Dock policy
- derive default retention from Dock entry visibility while preserving explicit snapshot metadata
- keep explicitly retained apps visible after their last window closes
- document the workspace app Dock lifecycle

## Validation

- pnpm check:changed (11 lanes)
- pre-push pnpm check:changed -- --push-ready (12 lanes)
- focused Dock projection and retention tests: 7 passed
- pnpm perf:agent-gui -- --scenario workbench-window-lifecycle

## Risk

Low. Existing explicit retained true/false metadata remains authoritative; the behavior change only affects installed apps without an explicit Dock choice.

### DIFF
--- a/apps/desktop/src/renderer/src/features/workspace-app-center/services/internal/workspaceAppCenterContribution.tsx
+++ b/apps/desktop/src/renderer/src/features/workspace-app-center/services/internal/workspaceAppCenterContribution.tsx
@@ -158,7 +158,8 @@ export function createWorkspaceAppCenterDockEntries(input: {
           clickBehavior: projection.clickBehavior,
           index,
           launchEnabled: projection.launchEnabled,
-          state: projection.state
+          state: projection.state,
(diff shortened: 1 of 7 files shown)


jomeswang:
Additional manual dev startup validation completed on the MR worktree:

- `make dev-gui` completed prerequisite checks and generated the builtin workspace app
- managed `tuttid`, CLI, Electron main/preload, and renderer warmup all completed successfully
- renderer dev server listened on `127.0.0.1:5173`
- `tutti-dev status` returned `service: tuttid` / `status: ok`
- Electron main, renderer, and managed daemon remained alive with no startup errors
- the dev process tree was stopped cleanly after validation

Only the existing Vite dynamic/static import advisory messages were emitted.

PR #2575, opened 2026-08-22
feat(analytics): track Agent, model, and Tutti Mode adoption

devRickyyy:
## Summary

- report confirmed Workspace Agent and Model Plan configuration changes plus completed Model Plan detection outcomes from tuttid
- add agent and model configuration-source dimensions to canonical root user-Turn terminal events
- attribute actual Tutti Mode use from the immutable, durably accepted Turn snapshot and report durable activation transitions
- keep all new analytics Tutti-only; shared Agent Host consumers such as TSH/VN do not receive these contracts

## Measurement

- Custom Agent adoption: agent_config_source = workspace_agent
- Custom model adoption: model_config_source = model_plan
- Tutti Mode intent: tutti_mode.activation_changed
- Tutti Mode actual use: tutti_mode_state = active on terminal Turn events
- configuration and detection funnels use bounded enums and exclude Agent/Plan names, IDs, endpoints, model names, credentials, preferences, and raw provider errors

## Validation

- pnpm check:changed (6 relevant lanes passed)
- pnpm check:changed -- --push-ready (7 lanes passed)
- focused Workspace Agent, Model Plan, Tutti Mode activation, terminal analytics, SQLite integration, and wiring tests passed
- git diff --check

## Compatibility

- Windows impact: none; the change is platform-neutral reporter and query logic with no path, process, shell, or filesystem behavior
- TSH/VN: unaffected because the reporter wiring and event ownership remain under services/tuttid

## Documentation

- updated docs/architecture/analytics-tracking.md with event contracts, privacy boundaries, and recommended DataFinder adoption metrics

### DIFF
--- a/docs/architecture/analytics-tracking.md
+++ b/docs/architecture/analytics-tracking.md
@@ -125,6 +125,97 @@ lifecycle writes. Non-AgentGUI prompt-session integrations keep their explicit
 tracker because they call the activity service without entering the shared
 engine.
 
+### Agent configuration usage dimensions
+
+The canonical root user-turn terminal events `agent.turn_completed`,
+`agent.turn_failed`, and `agent.turn_cancelled` carry three low-cardinality
+configuration and feature-use dimensions:
+
+| Param                 | Values                                       | Meaning                                              |
+| --------------------- | -------------------------------------------- | ---------------------------------------------------- |
+| `agent_config_source` | `workspace_agent`, `agent_target`, `unknown` | Whether the turn used a user-defined workspace Agent |
+| `model_config_source` | `model_plan`, `provider_native`, `unknown`   | Whether the turn used a user-defined Model Plan      |
+| `tutti_mode_state`    | `active`, `inactive`, `unknown`              | Whether Tutti Mode was active for that accepted Turn |
+
+The daemon resolves the two configuration-source dimensions at terminal-event delivery from the
+canonical session and its immutable `sessionRuntimeSnapshot`. It must not infer
+Model Plan use from a model name or provider, and it must not report Agent IDs,
+Model Plan IDs, names, endpoints, or credentials. Missing legacy sessions and
+invalid snapshots keep the terminal event but report `unknown`.
+
+The daemon resolves `tutti_mode_state` from the immutable Turn snapshot only
+after that snapshot has been durably accepted by the provider-dispatch path.
+It must not read the session's current activation when the Turn settles: the
+user may have changed the badge while the Turn was running. Missing,
+unaccepted, or unreadable snapshots report `unknown`. Analytics never includes
+activation/revision IDs or the effect and speed preferences.
+
+These events cover root, non-backfilled user-prompt turns with durable submit
+provenance. They exclude child/subagent turns, goal or automation turns, and
+provider-initiated turns. Recommended DataFinder metrics are:
+
+- Custom Agent user adoption: distinct DataFinder user identity with
+  `agent_config_source=workspace_agent`, divided by distinct identities whose
+  source is either `workspace_agent` or `agent_target`
+- Custom model user adoption: distinct DataFinder user identity with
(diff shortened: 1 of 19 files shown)


devRickyyy:
## Code Review

**Verdict: approve with notes.** No blocking issues found. One change requested before merge (finding 1), plus one cheap test case.

### Strengths

- **Attribution correctness is thought through.** `tutti_mode_state` reads the per-turn snapshot only after it's durably accepted, deliberately avoiding the settle-time race where the user flips the badge mid-turn. The "immutable snapshot" premise for the config-source dimensions checks out: `sessionRuntimeSnapshotContextKey` is written exclusively on the session-create path (`service_create.go:170,245`), so a delivery-time read reflects launch-time configuration. The check-accepted-then-read ordering is safe since acceptance is monotonic.
- **Privacy is enforced in depth, not just documented.** Params are bounded enums; `Build` re-normalizes all three dimensions to `unknown` on unrecognized input, so even a future caller passing a raw ID can't leak it; detection failure reasons go through an explicit allowlist (`normalizedDetectionFailureReason`); the activation event is gated by `IsUserStateSource`, which bounds `source` to `slash_command`/`badge_remove`. Every test asserts forbidden keys are absent.
- **Commit-then-report ordering is correct everywhere.** All `reportConfigurationChanged` calls sit after successful store writes; validation and store failures emit nothing, with tests for both. The two subtle gates verified: `resetComposerModel := plan.Enabled != enabled` (`modelplan/service.go:202`) is a true state-change gate, so the semantic-no-op suppression is real; and `DuplicatePlan` forces `Enabled = false`, so the test's `enabled` transition is genuine.
- **Graceful degradation matches the doc**: missing legacy sessions, unaccepted/missing snapshots, and read errors all keep the terminal event and report `unknown` — never dropped, never guessed. The `p.repo.(tuttiModeTurnSnapshotReader)` soft-wiring matches the existing `turnTerminalAnalyticsStore` pattern in the same file, and the SQLite integration test exercises the real store so the assertion can't silently rot.
- TSH/VN isolation claim verified: the wiring lives in `services/tuttid/wiring.go` and none of the touched services are shared with Agent Host consumers.

### Findings

**1. `Detect`: a completed detection vanishes if the snapshot persist fails** — `services/tuttid/service/modelplan/detection.go` (low severity, requested)

For saved-scope detections, `PutModelPlan` failure returns before `reportDetectionCompleted` runs. The detection *did* complete (the doc's contract is "detection attempt completed"); only its persistence failed. As written, saved-scope funnel data silently undercounts on store errors, and the doc's store-failure rule is stated for configuration funnels, not detection. Either emit before the persist step (the outcome is already final in `snapshot`) or add one doc sentence saying persist-failed detections are excluded.

**2. Test coverage gaps** (minor)

- No `disabled` action case — `TestServiceReportsCommittedModelPlanConfigurationChanges` covers `created/updated/duplicated/enabled/deleted` but never a true enabled→disabled transition.
- No `sessionErr` case in the config-sources table test, though the stub supports it (`sessionErr` field is added but never set).
- No test pinning `Detect`'s behavior when the saved-plan write fails (relevant to finding 1 — whichever behavior you pick, pin it).

**3. Nits** (optional)

- `IsTuttiModeTurnSnapshotAccepted` + `GetTuttiModeTurnSnapshot` is two SQLite round-trips that could be one query returning the snapshot with its accepted flag. Safe as-is.
- `recordingAnalyticsReporter` is now copy-pasted into three test packages; at three copies a shared test helper is justified.
- `DeletePlan` now reads the plan before the references check purely for analytics params; a concurrent update between read and delete reports slightly stale `protocol`/`template_kind`. Not worth fixing, just noting the window exists.

**4. Performance** — the terminal path gains up to three extra SQLite reads per settled root turn (GetSession + accepted-check + snapshot read). At turn-settle frequency with the 128-item drain limit this is negligible; no action needed.

---

PR #2578, opened 2026-08-22
feat(presence): report account device heartbeats

chovy-ai:
## Summary

Desktop clients need to report account-level device liveness independently of rooms and the disabled Mobile Remote feature. This adds a daemon-owned presence worker that registers the existing stable device identity, opens and immediately activates a server lease, and renews it at the server-provided interval.

The worker starts after the tuttid listener is ready or login completes, closes its exact lease before logout and during daemon shutdown, waits for a changed account cookie after `401`, reopens immediately after an expired lease, and uses bounded exponential backoff for transport or server failures. Device metadata continues to use the existing current-device endpoint and heartbeat requests contain no client clock or generic version field.

The control-plane adapter remains platform-neutral, forwards PPE headers, uses a five-second request timeout, and supports `TUTTI_DESKTOP_CONTROL_PLANE_BASE_URL` with the legacy Mobile Control Plane override as a compatibility fallback.

Depends on tutti-lab/tutti-server#42.

## Verification

- targeted lifecycle tests prove immediate activation, idempotent start, exact latest-lease close, expired-lease reopen, and waiting for a changed session after `401`
- HTTP contract tests prove metadata-only registration plus open/heartbeat/close paths and error preservation
- account logout ordering and wiring configuration tests pass
- targeted packages pass race detection and `go vet`
- `devicepresence` and `account` test binaries cross-compile for Windows amd64

The broader `go test ./service/...` run still hits pre-existing Codex subprocess/model-catalog timeout failures; all packages and tests changed by this PR pass.

## Fallback Three Questions

- Source: device sleep, process suspension, network loss, and transient control-plane failures can delay or lose heartbeats until the server lease expires.
- Why can it not be fixed at that source? These are external process/network conditions with no single reliable cross-platform event; Redis lease expiry must remain authoritative when the daemon cannot run.
- Removal condition: the retry/timer fallback can be removed only if account presence moves to another server-authoritative transport that provides equivalent expiry, reconnect, and fencing semantics.

## Checklist

- [x] If this PR changes agent lifecycle semantics (session/turn/goal/runtime-operation creation, sendability, terminal state, or recovery): the change lives in `packages/agent/host` and adds a `packages/agent/host/conformance` scenario, not in a `tuttid`/tsh adapter. This PR does not change agent lifecycle semantics.
- [x] I updated all README/CONTRIBUTING language variants when changing their English source. No README/CONTRIBUTING source changed.


chovy-ai:
Superseded by #2603, which includes the account device Presence heartbeat lifecycle together with the shared AccountRealtime and room-scoped user Presence client. Closing this earlier standalone implementation to avoid merging duplicate, diverged code.

PR #2624, opened 2026-08-27
fix(agent): preserve managed runtime for auth

jomeswang:
## Summary
- resolve Tutti Agent auth through the same provider command and managed Node environment as sessions and model discovery
- wire the unified bootstrap into browser-login readiness, model catalog refresh, and session preparation
- add sanitized process diagnostics and troubleshooting guidance

## Root cause
The auth bootstrap reused only the Tutti Agent launcher path and executed it with the daemon ambient environment. The npm launcher uses /usr/bin/env node, so login failed when Node existed only in the Tutti managed runtime even though account login had succeeded.

## Validation
- pnpm check:changed
- pre-push check:changed -- --push-ready
- go build ./... in services/tuttid
- GOOS=windows GOARCH=amd64 test compilation for service/tuttiagent and service/agent
- POSIX integration test with no system Node or tutti-agent on ambient PATH

## Windows impact
The shared resolver and environment contract are platform-neutral and have cross-platform unit coverage. Windows test binaries compile successfully; the executable shebang integration remains POSIX-only by design.

## Host boundary
This changes daemon-side provider auth composition only. Session lifecycle and provider protocol semantics are unchanged.

### DIFF
--- a/docs/conventions/troubleshooting/README.md
+++ b/docs/conventions/troubleshooting/README.md
@@ -28,8 +28,9 @@ Use the focused runtime index or open one area directly:
   Also covers focus-driven provider CLI scans, repeated Extension Target version
   probes, Windows managed-runtime adoption sharing violations, optional Provider
   absence misclassified as an environment failure, extension release refresh
-  delaying daemon startup, repeated Hermes helper downloads in isolated session
-  homes, and CPU spikes.
+  delaying daemon startup, Tutti Agent browser login that loses the managed Node
+  environment, repeated Hermes helper downloads in isolated session homes, and
+  CPU spikes.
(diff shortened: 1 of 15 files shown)


jomeswang:
macOS fresh-install E2E passed on the validation workflow: https://github.com/tutti-os/tutti/actions/runs/33052252958

Validated against the PR commit in an unsigned release DMG:
- x64, arm64, and universal DMGs built and smoke-installed successfully.
- Universal Tutti.app launched with an isolated state directory and a system-only PATH (no GitHub Actions Node or pre-existing tutti-agent on PATH).
- The installed desktop app made `local:tutti-agent` available and installed its executable.
- Cookie-based auth bootstrap completed successfully (`tutti_agent.auth_login.process_completed`, 192 ms).
- A real Tutti Agent session completed and returned the expected marker.

The temporary GitHub Environment/secret used by the E2E was deleted after the run.

